PR #1, opened 2020-12-03
Add missing setter from std vector

buschbapti:
This PR add the `set_velocities`, `set_accelerations` and `set_torques` functions with `std::vector<double>` as input parameter. `set_positions` was already implemented so this effectively just add the missing functions.

### DIFF
--- a/source/lib/state_representation/include/state_representation/Robot/JointState.hpp
+++ b/source/lib/state_representation/include/state_representation/Robot/JointState.hpp
@@ -93,7 +93,7 @@ namespace StateRepresentation
 		void set_positions(const Eigen::VectorXd& positions);
 
 		/**
-	 	 * @brief Setter of the posistions attribute
+	 	 * @brief Setter of the posistions from std vector
 	     */
 		void set_positions(const std::vector<double>& positions);
 
@@ -108,6 +108,11 @@ namespace StateRepresentation
 		void set_velocities(const Eigen::VectorXd& velocities);
 
 		/**
+	 	 * @brief Setter of the velocities from std vector
+	     */
+		void set_velocities(const std::vector<double>& velocities);
+
+		/**
 	 	 * @brief Getter of the accelerations attribute
 	     */
 		const Eigen::VectorXd& get_accelerations() const;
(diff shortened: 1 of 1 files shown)


eeberhard (commented):
It's a nice addition and very useful, but it also introduces a lot of code duplication. In this instance it would seem to make sense to have a private function that takes a reference to the class member variables (either position, velocity, acceleration or torque) and the vector to map. Each setter could then just wrap the generic function. If not for this->set_filled() it could even be a static function. 

```
(private) inline void JointState::set_state_variable(Eigen::VectorXd& state_variable, std::vector<double>& data) {
  if(data.size() != this->get_size()) throw IncompatibleSizeException("Input vector is of incorrect size: expected " + std::to_string(this->get_size()) + ", given " + std::to_string(data.size()));
  this->set_filled();
  state_variable = Eigen::VectorXd::Map(data.data(), data.size());
}
```
Generally copying code should be avoided since it will cause more effort to maintain later on

source/lib/state_representation/include/state_representation/Robot/JointState.hpp
```diff
@@ -301,6 +316,13 @@ namespace StateRepresentation
 		this->velocities = velocities;
 	}
 
+	inline void JointState::set_velocities(const std::vector<double>& velocities)
```
eeberhard, line 319: Here's an example where your local variable is shadowing the class member variable, e.g. overriding the symbol `velocities`. Your consistent use of `this->` makes this still readable so I think it's fine, but really shadowing member variables should be avoided where possible.

buschbapti: Yes in state_representation it was not done yet but everywhere else class member variables are now underscored to resolve this issue. This standard should be enforced in state_representation as well

buschbapti:
Following  @eeberhard review I have refactored the setters. I went a bit further and removed code duplication as much as possible. One test is failing but I have checked it was failing before those modifications. So we have to investigate there.

eeberhard (approved):
Looks good! Thanks for putting in the generic setter and reformatting, it cleans it up quite a bit.

domire8 (approved):
very nice improvement of the code

PR #2, opened 2020-12-18
Normal gain member for CircleDS

eeberhard:
* Add member normal_gain_ to Circular DS class

* Rename member gain_ to planar_gain_

* Distinguish between planar and normal gain in the main
evaluation: normal gain applies to the difference in local
Z axis, while planar gain is in the local XY.

* Update setters and getters (set_gain function sets both
gain members to the same value, so default  behaviour is unaffected)

* Reformat and fix some typos in docstrings

-- I would like this feature because I might want to treat normal gain differently than planar gain, for example to have a kind of "cylinder" response. In particular I want to control vertical gain externally (for adapting contact force), while tracking the planar gain much more closely.

Tip: hide whitespace changes when reviewing the changes

### DIFF
--- a/source/lib/dynamical_systems/src/Circular.cpp
+++ b/source/lib/dynamical_systems/src/Circular.cpp
@@ -1,59 +1,65 @@
 #include "dynamical_systems/Circular.hpp"
 
-namespace DynamicalSystems
-{
-	Circular::Circular(const StateRepresentation::CartesianState& center, double radius, double gain, double circular_velocity):
-	DynamicalSystem(StateRepresentation::CartesianPose::Identity(center.get_reference_frame())),
-	limit_cycle_(std::make_shared<StateRepresentation::Parameter<StateRepresentation::Ellipsoid>>("limit_cycle", StateRepresentation::Ellipsoid(center.get_name(), center.get_reference_frame()))),
-	gain_(std::make_shared<StateRepresentation::Parameter<double>>("gain", gain)),
-	circular_velocity_(std::make_shared<StateRepresentation::Parameter<double>>("circular_velocity", circular_velocity))
-	{
-		this->limit_cycle_->get_value().set_center_state(center);
-		this->limit_cycle_->get_value().set_axis_lengths({radius, radius});
-	}
+namespace DynamicalSystems {
+Circular::Circular(const StateRepresentation::CartesianState& center,
+                   double radius,
+                   double gain,
+                   double circular_velocity) :
+    DynamicalSystem(StateRepresentation::CartesianPose::Identity(center.get_reference_frame())),
+    limit_cycle_(std::make_shared<StateRepresentation::Parameter<StateRepresentation::Ellipsoid>>("limit_cycle",
+                                                                                                  StateRepresentation::Ellipsoid(
+                                                                                                      center.get_name(),
+                                                                                                      center.get_reference_frame()))),
+    planar_gain_(std::make_shared<StateRepresentation::Parameter<double>>("gain", gain)),
+    circular_velocity_(std::make_shared<StateRepresentation::Parameter<double>>("circular_velocity",
+                                                                                circular_velocity)) {
+  this->limit_cycle_->get_value().set_center_state(center);
+  this->limit_cycle_->get_value().set_axis_lengths({radius, radius});
+}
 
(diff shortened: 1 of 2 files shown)


source/lib/dynamical_systems/src/Circular.cpp
```diff
@@ -1,59 +1,65 @@
...
-		return param_list;
-	}
+namespace DynamicalSystems {
+Circular::Circular(const StateRepresentation::CartesianState& center,
+                   double radius,
+                   double gain,
+                   double circular_velocity) :
+    DynamicalSystem(StateRepresentation::CartesianPose::Identity(center.get_reference_frame())),
+    limit_cycle_(std::make_shared<StateRepresentation::Parameter<StateRepresentation::Ellipsoid>>("limit_cycle",
+                                                                                                  StateRepresentation::Ellipsoid(
```
buschbapti, line 10: incosistent ident

source/lib/dynamical_systems/src/Circular.cpp
```diff
@@ -1,59 +1,65 @@
...
+    planar_gain_(std::make_shared<StateRepresentation::Parameter<double>>("gain", gain)),
+    circular_velocity_(std::make_shared<StateRepresentation::Parameter<double>>("circular_velocity",
+                                                                                circular_velocity)) {
+  this->limit_cycle_->get_value().set_center_state(center);
+  this->limit_cycle_->get_value().set_axis_lengths({radius, radius});
+}
+
+Circular::Circular(const StateRepresentation::Ellipsoid& limit_cycle, double gain, double circular_velocity) :
+    DynamicalSystem(StateRepresentation::CartesianPose::Identity(limit_cycle.get_center_pose().get_reference_frame())),
+    limit_cycle_(std::make_shared<StateRepresentation::Parameter<StateRepresentation::Ellipsoid>>(" limit_cycle",
```
buschbapti, line 23: I don't think the space in the limit_cycle name should be here. If it was already then it is a typo and should be corrected

source/lib/dynamical_systems/src/Circular.cpp
```diff
@@ -1,59 +1,65 @@
...
+  StateRepresentation::CartesianTwist velocity(pose.get_name(), pose.get_reference_frame());
+  Eigen::Vector3d linear_velocity;
+  linear_velocity(2) = -this->get_normal_gain() * pose.get_position()(2);
+
+  std::vector<double> radiuses = this->get_radiuses();
+
+  double a2 = radiuses[0] * radiuses[0];
+  double b2 = radiuses[1] * radiuses[1];
+  double dradius = -this->get_planar_gain() * radiuses[0] * radiuses[1]
+      * (pose.get_position()[0] * pose.get_position()[0] / a2 + pose.get_position()[1] * pose.get_position()[1] / b2
```
buschbapti, line 44: this identation makes it hard to read

source/lib/dynamical_systems/src/Circular.cpp
```diff
@@ -11,6 +11,7 @@ Circular::Circular(const StateRepresentation::CartesianState& center,
                                                                                                       center.get_name(),
                                                                                                       center.get_reference_frame()))),
     planar_gain_(std::make_shared<StateRepresentation::Parameter<double>>("gain", gain)),
```
buschbapti, line 13: gain parameter should be renamed to planar_gain parameter then. Parameters are mostly called from command line so I am ok with changing them. It should not break any behaviors

buschbapti (approved):
I am good with the idea and the code. Mostly some identation to correct and good to go.

domire8 (commented):
good for me after you did the changes requested by Baptiste.

We might have to discuss the indentation problems

source/lib/dynamical_systems/src/Circular.cpp
eeberhard: ok, this is what i was unsure about so I didn't want to break existing behaviours. good to know, will fix.

source/lib/dynamical_systems/src/Circular.cpp
eeberhard: I already found it hard to read before, it's a lot of calculation for one line - but I agree, I didn't catch what the autoformatter did here with the line break

PR #4, opened 2021-02-27
Correct indentation and snake_case namespaces

buschbapti:
It is crazy PR time again. As always this is just an indentation refactor. Also I have changed all the namespaces to adopt snake_case naming convention and remove some of the nested ones.


source/packages/modulo_core/tests/testMovingReferenceFrame.cpp
```diff
@@ -1,202 +1,173 @@
-#include "modulo_core/Cell.hpp"
-#include "state_representation/Space/Cartesian/CartesianState.hpp"
-#include "state_representation/Space/Cartesian/CartesianPose.hpp"
-#include "state_representation/Space/Cartesian/CartesianTwist.hpp"
 #include "dynamical_systems/Circular.hpp"
```
domire8, line 1: If you want to be very consistent, the state_rep and DS includes should now be treated as external libs (with `<>` and after the local header includes)

source/packages/modulo_core/include/modulo_core/Exceptions/NotImplementedExcpetion.hpp
```diff
@@ -1,23 +1,11 @@
...
-	{
-		class NotImplementedException : public std::logic_error
-		{
-		public:
-			explicit NotImplementedException(const std::string& msg) : logic_error(msg)
-			{};
-		};
-	}
-}
+namespace modulo::core::exceptions {
```
domire8, line 6: Getting rid of these nested namespaces is so satisfying :smile:

domire8 (approved):
Great, the formatting and namespace changes are very nice! I trust that you didn't change the code otherwise and I hope that one day I can use this code for real

source/packages/modulo_core/tests/testMovingReferenceFrame.cpp
buschbapti: that is a good point yes that should be there as well. I will change that

source/packages/modulo_core/include/modulo_core/Exceptions/NotImplementedExcpetion.hpp
buschbapti: ah ah yes indeed!

PR #5, opened 2021-03-29
Use controller libraries on release branch

buschbapti:
Sorry for this massive PR again. This propagates all the changes we have made in `control_libraries`. Everything is compiling fine, I don't think functionalities have been broken since it is mainly namespaces that have changed.


Dockerfile
```diff
@@ -2,51 +2,10 @@ ARG ROS_VERSION=foxy
...
-RUN mkdir build && cd build && cmake .. && make -j && sudo make install
-
-# import previously downloaded packages
-RUN mkdir -p ${HOME}/modulo_lib
-WORKDIR ${HOME}/modulo_lib/
-COPY --chown=${USER} ./source/lib/ .
-# build packages and libraries
-RUN sh build.sh
+# import control library packages
+RUN git clone -b main --single-branch https://github.com/epfl-lasa/control_libraries.git
```
domire8, line 6: Wouldn't it be better to check out the tag? I know it's essentially the same because we keep main clean but it would be clearer that you're using a certain version here?

Dockerfile
```diff
@@ -56,6 +15,8 @@ RUN /bin/bash -c "source /opt/ros/$ROS_DISTRO/setup.bash; colcon build"
 # change to the home root
 WORKDIR ${HOME}
 
+ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/openrobots/lib/
```
domire8, line 18: Is this needed again? its already done in the install script

domire8 (approved):
Looks good to me

Dockerfile
buschbapti: This is related to the problem that the libraries must be installed as `root` in `modulo`. When using `sudo` the environment variables are not set. This is the only hack I have found to make the compilation work.

domire8: why are you not root here? I guess because you base on your own ros2_ws image where you created a user at the end?

buschbapti: yes

source/packages/modulo_core/src/Communication/MessagePassing/WriteStateConversion.cpp
```diff
@@ -157,34 +157,27 @@ void write_msg(geometry_msgs::msg::Twist& msg, const StateRepresentation::DualQu
...
 
-void write_msg(tf2_msgs::msg::TFMessage& msg, const StateRepresentation::CartesianState& state, const rclcpp::Time& time) {
+void write_msg(tf2_msgs::msg::TFMessage& msg, const state_representation::CartesianState& state, const rclcpp::Time& time) {
   if (state.is_empty()) throw EmptyStateException(state.get_name() + " state is empty while attempting to publish it");
   geometry_msgs::msg::TransformStamped transform;
   write_msg(transform, state, time);
   msg.transforms.push_back(transform);
 }
 
-void write_msg(std_msgs::msg::Float64MultiArray& msg, const StateRepresentation::CartesianTwist& state, const rclcpp::Time&) {
```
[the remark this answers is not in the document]
buschbapti: Ah no actually, thanks to the `data()` and `to_std_vector()` that are in all classes of `state_representation` (`CartesianState`, `JointState`, `Parameters`) we don't need to use a dedicated function anymore and can use the template version.

PR #6, opened 2021-05-28
Correct installation of control_libraries

buschbapti:
Due to the recent changes of `control_libraries`, the install script now requires the `-y` option. I am also changing for the develop branch as we are here tracking develop I guess. When doing release in here we should then track a released branch of `control_libraries`


domire8 (approved):
Sounds good!

eeberhard (approved):

PR #8, opened 2021-06-03
Fix Jacobian setters in ReadStateConversion

eeberhard:
* Recent removal of the Jacobian::set_rows and ::set_cols
methods from the control_libraries development stream means
that these lines have been removed. Now, if the passed
Jacobian does not have the right size, it will
throw an IncompatibleSizeException.

If it's undesirable to throw an exception within the read message, the alternative is to construct and swap a new Jacobian object with the corresponding robot name from the original state and everything else from the message state. I don't know what you prefer here - lightweight code that can raise exceptions, or slightly heavier code that is exception safe. I feel like the former is preferable in the message layer.


buschbapti (approved):
Yes I am happy with this. It makes sense to me that it can raise exceptions

PR #9, opened 2021-06-09
Solve issues to use ros galactic

buschbapti:
We can now change to ROS Galactic and there was some issues with part of the code. Specifically one header was not included. The rest was simple string issues.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-ROS_VERSION=foxy
+ROS_VERSION=galactic
 
 IMAGE_NAME=aica-technology/modulo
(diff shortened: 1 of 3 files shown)


domire8 (approved):
Wow already galactic! They move almost as fast as us with control_libraries :smile: I don't see an issue here, all good

eeberhard (approved):
The changes herein are compatible with foxy, so even if we aren't using galactic for now it's worth merging this.

I just rebased this onto develop and reset the build script version back to foxy.

buschbapti:
Yes seems good. I will then merge this.

PR #12, opened 2021-06-24
Fix periodic call exiting before configure

buschbapti:
The `periodic_call` functions are defined in the same way than the normal `step` function with a `run` function that loop while the node is `configured`. This effectively means that the `add_periodic_call` function needs be put in the `on_configure` function of the node. Problem was that the`on_configure` function of the derived nodes is called prior the setting of the `configured_` boolean. This means that the `run_periodic_call` function was exiting as soon as it was added. Simple solution is to set the `configured_` boolean prior the `on_configure` call. In case the node fails to configure anyway, the reset function is called which terminates all threads and reset all handlers of threads and publishers. So I think it is actually rather safe to do like this.

### DIFF
--- a/source/packages/modulo_core/src/Cell.cpp
+++ b/source/packages/modulo_core/src/Cell.cpp
@@ -327,7 +327,6 @@ const state_representation::CartesianPose Cell::lookup_transform(const std::stri
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Cell::on_configure(const rclcpp_lifecycle::State&) {
   RCUTILS_LOG_INFO_NAMED(get_name(), "on_configure() is called.");
-  this->active_ = false;
   this->configured_ = true;
   // call the proxy on_configure function
   if (!this->on_configure()) {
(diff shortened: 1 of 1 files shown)


source/packages/modulo_core/src/Cell.cpp
```diff
@@ -327,15 +327,14 @@ const state_representation::CartesianPose Cell::lookup_transform(const std::stri
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Cell::on_configure(const rclcpp_lifecycle::State&) {
   RCUTILS_LOG_INFO_NAMED(get_name(), "on_configure() is called.");
+  this->active_ = false;
```
eeberhard, line 330: just as a curiosity, what happens if we call configure() when a node has already been activated?

buschbapti: You can't actually. ROS won't let you do it. So in fact this line is problably not needed

eeberhard: I think it might be good to remove `this->active_ = false;` in that case

eeberhard (approved):
Simple enough and makes sense

PR #13, opened 2021-07-14
Add NodeOptions constructor to Cell

eeberhard:
* Add constructor declaration and
definition for Cell to take a
NodeOptions structure and parse
the name and period. This is
so that the Cell can be configured
as a component.

* Add utilities header to make parsing
NodeOptions more convenient. This utility
can be used by derived classes outside
of modulo core.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,8 @@
 **/*#*
 **/__pycache__/**
 **/.cache/**
+
+.idea
+cmake-build-*
+
 /source/lib/state_representation/build/*
--- a/source/packages/modulo_core/include/modulo_core/Cell.hpp
+++ b/source/packages/modulo_core/include/modulo_core/Cell.hpp
@@ -26,14 +26,14 @@ using namespace std::chrono_literals;
 namespace modulo::core {
 /**
  * @class Cell
- * @brief Abstract class to define a Call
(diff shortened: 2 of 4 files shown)


buschbapti (approved):
This looks good an handy!

buschbapti:
I see there are some docstrings missing. Otherwise seems good like this. Do you know if the namespace is something used out of the box or should the cell actually use it and take it as argument?

eeberhard:
> Do you know if the namespace is something used out of the box or should the cell actually use it and take it as argument?

I added the namespace parsing in case we want to extract the namespace from the NodeOption outside of construction for whatever reason. When a node or a lifecycle node is constructed, it will actually use the remapping argument if it is present to qualify the full node name. Now that you mention it though, the Cell will not support remapping if I don't also pass the option struct through to the underlying constructor... I guess I'll add that.

eeberhard:
> I see there are some docstrings missing.

@buschbapti could you point out where docstrings are missing?

buschbapti:
Sorry my bad it is alright

buschbapti:
I was just looking at the `@param` or `@return` that do not have description. But maybe they did need one. I just always put something even if a bit redundant.

buschbapti (approved):

PR #14, opened 2021-07-14
Use wall timer instead of threads

buschbapti:
When discussing with @eeberhard I realized that using thread in the `Cell` class was actually breaking ros2 design that rely on a specific pattern called wall timer. This is used to add periodic publishing functions, but in fact it is also made to add any type of periodic call and works fine to handle the `step` function.

I have therefore removed all threads to switch to this wall timer and everything behave smoothly. I initially thought that those timer functions would not be called when the node is `deactivated` but that is not the case. In a sense this is actually good for us we can encapsulate the callback functions in another function that checks for node activation and also handle the mutex guard. However, it also means that this will probably not solve the throttling issue when too many nodes are spawned.

### DIFF
--- a/source/packages/modulo_core/include/modulo_core/Cell.hpp
+++ b/source/packages/modulo_core/include/modulo_core/Cell.hpp
@@ -475,8 +475,7 @@ Cell::Cell(const std::string& node_name, const std::chrono::duration<int64_t, Du
                                                                                                                               mutex_(std::make_shared<std::mutex>()),
                                                                                                                               period_(period) {
   // add the update parameters call
-  std::function<void(void)> update_parameters_fnc = std::bind(&Cell::update_parameters, this);
-  this->update_parameters_timer_ = this->create_wall_timer(this->period_, update_parameters_fnc);
+  this->update_parameters_timer_ = this->create_wall_timer(this->period_, [this] { this->update_parameters(); });
 }
 
 inline const std::map<std::string, std::pair<std::shared_ptr<communication::CommunicationHandler>, bool>>& Cell::get_handlers() const {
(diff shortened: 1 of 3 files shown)


source/packages/modulo_core/include/modulo_core/Cell.hpp
```diff
@@ -478,7 +476,7 @@ Cell::Cell(const std::string& node_name, const std::chrono::duration<int64_t, Du
                                                                                                                               period_(period) {
   // add the update parameters call
   std::function<void(void)> update_parameters_fnc = std::bind(&Cell::update_parameters, this);
-  this->parameter_update_thread_ = std::thread(update_parameters_fnc);
+  this->update_parameters_timer_ = this->create_wall_timer(this->period_, update_parameters_fnc);
```
eeberhard, line 479: I think generally lambda functions are preferred over binding in modern usage. You would save yourself an intermediate variable too. Since these lines are being changed anyway, maybe you could update them here and in the other places (unless you disagree of course)
```suggestion
  this->update_parameters_timer_ = this->create_wall_timer(this->period_, [this] { this-> update_parameters(); });
```

eeberhard (approved):
Nice change. I got around to testing the build and run and it seems to work fine still. I particularly like that we can remove the thread->join steps in the reset.

source/packages/modulo_core/include/modulo_core/Cell.hpp
buschbapti: no sounds perfectly good. I didn't know about that syntax tbh

buschbapti:
Good I have changed the `bind` to lamda functions. This is so much cleaner.

In `testCartesian` I have just added a temporary call to a dummy printing function just for you to test that all works fine. In particular, I have tested the `deactivate -> cleanup -> configure -> activate` cycle to verify that the periodic calls are well destroyed and recreated without overlapping. Everything seems to work fine. When you validate it I will remove that dummy call (deleting the actual commit) before merging.

eeberhard:
Great stuff, I tested it and can confirm that the print happens as expected, only when the node is activated. Thanks for adding this demonstration to the PR to make things easy!

Happy for you to merge, then I'll rebase the updated NodeOptions PR after

PR #16, opened 2021-08-12
Build and push modulo image in CI

domire8:
Pretty self-explaining, one new thing is that I added an `if` condition to decide between pushing a tagged version linked to a release or a latest version based on a push on main or develop.


buschbapti (approved):
This seems good. Thanks!

PR #17, opened 2021-08-17
Use protobuf bindings for publisher subscriber

buschbapti:
Title says it all, this PR includes the prtobuf bindings to send encode state through ROS2 pub sub!


source/packages/modulo_core/include/modulo_core/Communication/EncodedState.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <std_msgs/msg/byte_multi_array.hpp>
+
+namespace modulo::core {
+	using EncodedState = std_msgs::msg::ByteMultiArray;
```
eeberhard, line 6: I prefer a `typedef` to a `using` declaration, although I think it's the same in the end 

```suggestion
	typedef std_msgs::msg::ByteMultiArray EncodedState;
```

source/packages/modulo_core/CMakeLists.txt
```diff
@@ -19,11 +19,13 @@ list(APPEND CMAKE_PREFIX_PATH /opt/openrobots)
 find_package(pinocchio REQUIRED)
 find_package(OsqpEigen REQUIRED)
 find_package(osqp REQUIRED)
+find_package(Protobuf REQUIRED)
```
eeberhard, line 22: As discussed, `find_library(protobuf REQUIRED)` will be sufficient as we only care about the runtime `libprotobuf.so` and not any of the additional package stuff (like cmake macros, etc)

source/packages/modulo_core/include/modulo_core/Communication/MessagePassing/ReadStateConversion.hpp
```diff
@@ -148,13 +151,25 @@ void read_msg(state_representation::DualQuaternionTwist& state, const geometry_m
...
 }
+
+/**
+ * @brief Convert a ROS std_msgs::msg::ByteMultiArray to a State using protobuf decoding
+ * @tparam a state_representation::State type object
+ * @param state The state to populate
+ * @param msg The ROS msg to read from
+ */
+template <typename T>
+void read_msg(T& state, const std_msgs::msg::ByteMultiArray& msg) {
```
eeberhard, line 171: For consistency maybe use `modulo::core::EncodedState` here, and in WriteState also... then you can include `Communication/EncodedState.hpp` instead of `std_msgs/msg/byte_multi_array.hpp`, and it perhaps stays a little easier to maintain

source/packages/modulo_core/tests/testCartesian.cpp
```diff
@@ -51,8 +51,8 @@ class ConsoleVisualizer : public modulo::core::Cell {
                                                                                                       desired_twist(std::make_shared<state_representation::CartesianTwist>("robot_test")) {}
 
   bool on_configure() {
-    this->add_subscription<geometry_msgs::msg::PoseStamped>("/robot_test/pose", this->robot_pose);
-    this->add_subscription<geometry_msgs::msg::TwistStamped>("/ds/desired_twist", this->desired_twist);
+    this->add_subscription<modulo::core::EncodedState>("/robot_test/pose", this->robot_pose);
```
eeberhard, line 54: It might be worth adding signatures for that don't even require the template. For example, `add_state_subscription` and `add_state_publisher` would internally call `add_subscription<modulo::core::EncodedState>` and so on. This makes things even easier for the end user of a Cell-derived class, which is what we ultimately want.

eeberhard (commented):
Awesome to see it coming together!

eeberhard (approved):

PR #18, opened 2021-08-17
Add Component class

buschbapti:
Add the `Component` class coming from `state_machine` (previously `Action`) stripping it down to the basic handling of predicates


eeberhard (approved):
Nice work with this! I think the received predicates and events are not needed by the downstream components in the state engine architecture, but they may be useful elsewhere, so it's good to have anyway.

PR #19, opened 2021-08-18
Refactor modulo by removing modulo_msgs

buschbapti:
As for #17, we can now use protobuf to directly send state based messages. Effectively this removes the need for custom messages such as the `Jacobian` one. This PR, therefore, remove all dependencies to `modulo_msgs` and cleans the folder architecture.

Folders were also not using snake case so I have changed that for consistency with our style guideline.


eeberhard (commented):
Seems good!

In order for this image to build in the CI (and therefore to use this downstream) we need to resolve the introduced error. This can be a separate PR but it needs to be resolved soon.
```
Step 4/20 : COPY --from=ghcr.io/epfl-lasa/control-libraries/development-dependencies:latest /usr/local/lib/libproto* /usr/local/lib
219
When using COPY with more than one source file, the destination must be a directory and end with a /
220
Error: Process completed with exit code 1.
```

eeberhard (approved):
Good to go! I will make the CI fix in a separate PR

PR #20, opened 2021-08-19
Fix docker COPY in CI

eeberhard:
The CI was failing with the new libproto copy command:
> When using COPY with more than one source file, the destination must be a directory and end with a /

I added a temporary CI build on PR synchronise to test this. I will remove it once the image build.

### DIFF
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -9,6 +9,7 @@ on:
   release:
     types: [published]
   workflow_dispatch:
+  pull_request:
 
 jobs:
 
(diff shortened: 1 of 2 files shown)


eeberhard:
As can be seen by commit [348594e](https://github.com/epfl-lasa/modulo/pull/20/commits/348594eb5aebde5b7b7efc7822a7f2cc0279e0bd), the simple change was enough to fix the CI issue. The CI dispatch was then reverted to normal

buschbapti (approved):
Good catch here! Tested and validated

PR #22, opened 2021-08-19
Fix the predicate publishing that was only happening at activation

buschbapti:
The whole predicate publishing was only happening when the node was actually activated. This fixes this issues by changing the scope of `always_activated` communication handlers. It also introduces the concept of `daemons` function calls that are always happening in the background even if the node is not configured (e.g. `update_parameters` function call.

### DIFF
--- a/source/modulo_core/include/modulo_core/Cell.hpp
+++ b/source/modulo_core/include/modulo_core/Cell.hpp
@@ -531,6 +531,11 @@ public:
 
   /**
    * @brief Function to add a periodic call to the function given in input
+   * @details This function is similar to the add_deamon one with
+   * the difference that the callback_function is only called when the node
+   * is active. The desired design is, therefore, that it should only be called
+   * in the on_configure function, as periodic_call will be cleaned when on_cleanup
+   * is called.
    * @param callback_function the function to call
    * @param period the period between two calls
    */
@@ -538,6 +543,10 @@ public:
 
   /**
    * @brief Function to daemonize the callback function given in input
+   * @details This function is similar to the add_periodic_call one with
+   * the difference that a daemon is always active, including when the node
+   * is not configured. The desired design is, therefore, that it should be
+   * called only in the constructor to avoid potential duplicates.
    * @param callback_function the function to call
    * @param period the period between two calls
    */
--- a/source/modulo_core/include/modulo_core/Cell.hpp
+++ b/source/modulo_core/include/modulo_core/Cell.hpp
@@ -42,10 +42,11 @@ private:
   bool active_;                                                                                          ///< boolean that informs that the node has been activated, i.e passed by the on_activate state
   bool shutdown_;                                                                                        ///< boolean that informs that the node has been shutdown, i.e passed by the on_shutdown state
   std::shared_ptr<rclcpp::TimerBase> update_parameters_timer_;                                           ///< reference to the timer for periodically that periodicall update the parameters
-  std::list<std::shared_ptr<rclcpp::TimerBase>> timers_;                                                 ///< reference to timers for periodically called functions
   std::chrono::nanoseconds period_;                                                                      ///< rate of the publisher functions in nanoseconds
   std::map<std::string, std::shared_ptr<state_representation::ParameterInterface>> parameters_;          ///< list for storing parameters
   std::map<std::string, std::pair<std::shared_ptr<communication::CommunicationHandler>, bool>> handlers_;///< map for storing publishers, subscriptions and tf
+  std::list<std::shared_ptr<rclcpp::TimerBase>> timers_;                                                 ///< reference to timers for periodically called functions
+  std::list<std::shared_ptr<rclcpp::TimerBase>> daemons_;                                                ///< reference to timers for daemonized functions
 
   /**
    * @brief Function to clear all publishers, subscriptions and services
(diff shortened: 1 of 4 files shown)


source/modulo_core/src/Component.cpp
```diff
@@ -28,20 +28,23 @@ void Component::add_predicate(const std::shared_ptr<state_representation::Predic
...
+  this->add_publisher<std_msgs::msg::Bool>("/predicates/" + std::string(this->get_name()) + "/" + predicate_name, predicate, true);
 }
 
 void Component::add_predicate(const std::string& predicate_name, const std::function<bool(void)>& predicate_function) {
   // crate a predicate with named prefixed with the action name
   auto predicate = std::make_shared<state_representation::Predicate>(predicate_name);
   this->add_predicate(predicate);
   // insert the predicate function in the list
-  this->predicate_functions_.push_back(std::make_pair(predicate_name, predicate_function));
+  this->predicate_functions_.insert(std::make_pair(predicate_name, predicate_function));
```
eeberhard, line 40: I'm tempted to say that `insert_or_assign` would be better here, so that we can overwrite the callback function for an existing predicate if needed

source/modulo_core/include/modulo_core/Cell.hpp
```diff
@@ -534,6 +535,13 @@ class Cell : public rclcpp_lifecycle::LifecycleNode {
    * @param period the period between two calls
    */
   void add_periodic_call(const std::function<void(void)>& callback_function, const std::chrono::nanoseconds& period);
+
+  /**
+   * @brief Function to daemonized the callback function given in input
```
eeberhard, line 540: This and the method above could benefit from a slight elaboration or even a @details tag,
to disambiguate it from the periodic call (that the callback function
is always evaluated, while the periodic call wraps the callback
in the run_periodic_call method to filter for the active state

Also
```suggestion
   * @brief Function to daemonize the callback function given in input
```

source/modulo_core/src/Cell.cpp
```diff
@@ -16,18 +16,25 @@ Cell::Cell(const rclcpp::NodeOptions& options) :
...
 
 Cell::~Cell() {
-  this->parameters_.clear();
   this->on_shutdown();
 }
 
 void Cell::reset() {
   this->active_ = false;
   this->configured_ = false;
-  this->handlers_.clear();
```
eeberhard, line 30: This makes it evident that handlers cannot be added in the constructor, since this would remove the interfaces when unconfiguring and reconfiguring a node. Regular handlers, even if they are "always active", will never work in an unconfigured state, since they must be added "on_configure".

Since Component predicate functions are now daemonized, they persist throughout the cleanup and reset. It makes sense now!

eeberhard (commented):
This is a good change overall and makes a lot of sense in context. I had just minor comments, and still want to test a couple of things with the example applications in the state engine as a consequence.

We eventually need good usage documentation, ideally in the style of [the ros2 lifecycle design article](https://design.ros2.org/articles/node_lifecycle.html), but that item is stuck in the backlog for now.

source/modulo_core/src/Cell.cpp
buschbapti: Yes exactly. Now with this PR you can add handlers in the constructor as well as long as you declare them `always_active`. This is actually how it is done in the `Component` class for predicates.

source/modulo_core/src/Component.cpp
buschbapti: Yes this seems fair. It could actually be the default behavior for all handlers in `Cell`

buschbapti: Ah well I am not sure actually. So the problem is we now directly assign the publisher. However, ros will throw an error if you assign a new publisher / subscriber on the same topic. So I actually think the proper behavior would be to throw an exception

buschbapti: I actually do that line 26 in the `add_predicate` function called the line above (38):

```cpp
if (this->predicates_.find(predicate_name) != this->predicates_.end()) {
    throw exceptions::PredicateAlreadyRegisteredException(predicate_name);
  }
```

eeberhard: I don't want to overwrite the handler (since removing and recreating the handler on the same interface / topic is not useful), rather just the callback function in this line

source/modulo_core/src/Cell.cpp
eeberhard: Ah yes perfect. In fact I wrote it wrong in my first comment which is how it was before this PR. Since now the reset() only erases non-always active handlers, it looks like it works as intended!

source/modulo_core/src/Component.cpp
buschbapti: I understand that. But this is not possible withing this call. If you do:

```cpp
add_predicate("pred", callbackA)
add_predicate("pred", callbackB)
```

It will throw a `PredicateAlreadyRegisteredException` coming from the `add_predicate(predicate)` call line 38.

buschbapti: This is resolved now by catching the exception and changing the call order

eeberhard (approved):
Awesome, the daemon predicates are now working along with any always active handlers. This change finally enables the state engine to correctly transition lifecycle nodes! Very exciting.

Tested with local changes equivalent to the most recent commit and seems to be working great downstream.

PR #23, opened 2021-09-01
Fix the removed from_std_vector

buschbapti:
Due to the latest push in `control_libraries` that removes `from_std_vector`, `modulo` was in a broken state. This PR fixes it by using `set_data`. It also introduces a new Dockerfile structure that allow to create a remote development server even if modulo is not compiling. This might need to change the CI handles the images.

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -28,5 +28,5 @@ COPY --chown=${USER} ./source/ ./src/
 # Clean image
 RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
 
-# add the key for robotpkg
+# Restore the key for robotpkg
 RUN sudo curl http://robotpkg.openrobots.org/packages/debian/robotpkg.key | sudo apt-key add -
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -21,12 +21,10 @@ RUN sudo ./install.sh
 ENV LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/opt/openrobots/lib/
 RUN sudo ldconfig
 
+WORKDIR ${HOME}/ros2_ws
 # copy sources and build ROS workspace with user permissions
 WORKDIR ${HOME}/ros2_ws/
 COPY --chown=${USER} ./source/ ./src/
-RUN /bin/bash -c "source /opt/ros/$ROS_DISTRO/setup.bash; colcon build"
-
-WORKDIR ${HOME}
 
 # Clean image
 RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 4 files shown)


domire8:
Can you elaborate why you need two Dockerfiles for that and not one with two stages?

Dockerfile.production
```diff
@@ -0,0 +1,5 @@
+ARG ROS_VERSION=foxy
+FROM aica-technology/modulo/development:${ROS_VERSION} as remote-development
+
+# build modulo
+RUN /bin/bash -c "source /opt/ros/$ROS_DISTRO/setup.bash; colcon build"
```
domire8, line 5: I personally prefer doing that instead of changing the user to ${USER} on line 10 in the development Dockerfile
```suggestion
RUN su ${USER} -c /bin/bash -c "source /opt/ros/$ROS_DISTRO/setup.bash; colcon build"
```

buschbapti:
This was actually recommended by @eeberhard for the CI. I am happy to change that for multistage if more appropriate

domire8:
> This was actually recommended by @eeberhard for the CI. I am happy to change that for multistage if more appropriate

Okay then my follow up question would be why we want to build and push an image that doesnt even build modulo..? Sorry, maybe I miss something again, modulo is not my strength :smile:

buschbapti:
Well this is simple to answer. Remember that clion needs to access a container through ssh right? How do you do that if your image fails at building because your code needs fixing as it was the case here? The only solution is to build an image that install everything but does not try to build it.

domire8:
Yeah obviously, I guess my question was rather why not just use a multi stage Dockerfile. But yes, if we were to build two packages in then two Dockerfiles are necessary.

However, IF we build and push an image with control libraries installed, I think we should do it in a dedicated place and for both ros and ros2, maybe even with the python bindings installed. Because that idea already crossed my mind several times. But if you think its faster and easier to do it this way in modulo, I can live with it.

eeberhard (approved):
The change to the actual logic is good. Regarding the dockerfile structure, I think it's quite good. 

I second dominic's comment to run the colcon build as the correct user.

Dockerfile.development
```diff
@@ -27,3 +27,6 @@ COPY --chown=${USER} ./source/ ./src/
 
 # Clean image
 RUN sudo apt-get clean && sudo rm -rf /var/lib/apt/lists/*
+
+# add the key for robotpkg
```
eeberhard, line 31:
```suggestion
# Restore the key for robotpkg
```

eeberhard (approved):
Nice catch with the apt key at the end, I forgot about that but it's definitely helpful.

I think we can merge this now to keep things running, and then in future reduce some CI steps by basing off of the proposed base image in https://github.com/aica-technology/docker-images/pull/13

PR #25, opened 2021-09-08
Use ros2-control-libraries image

buschbapti:
Now that the image combining ros2 and control-libraries has been pushed, it can be used as a base for modulo


domire8 (approved):
:+1: :rocket: :fire: :100:

eeberhard (approved):
I looked at #26 first and was about to suggest this change be done in that PR before I saw this one 😆 

Nice, I don't think the order of closing the PRs will matter. If this close before #26 then the CI will have to do less work and we save some bandwidth.

PR #27, opened 2021-09-20
Improve the script to build also the production image

buschbapti:
The current building process for the development and production image was not very convenient as it required to modify the script in case you wanted to build the production image. This PR introduces the `-p|--production` option to the script to build the production image. Note that it also build the development one first and uses the `-r` option if provided. This PR also correct the pulled image which was not changed to `ros2-contol-libraries`

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,23 +1,33 @@
 #!/bin/bash
 ROS_VERSION=foxy
 
-TARGET=development
+BUILD_PROD=false
 IMAGE_NAME=aica-technology/modulo
 
-BUILD_FLAGS=()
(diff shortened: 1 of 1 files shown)


domire8 (approved):

eeberhard (approved):
This makes a lot of sense, and both builds work fine for me locally. Would be nice to have a help message in the script on invalid arguments and with the -h flag as in the other scripts (see for example aica-docker) if you would like to add that.

domire8 (approved):

eeberhard (approved):

PR #28, opened 2021-09-22
Release/v1.0.0

eeberhard:
It's release time! 🚂 

Modulo hasn't had a release in 10 months. Since then, we've made a great deal of changes, the most fundamental being the separation of state_representation and related libraries from modulo into control-libraries repo. I didn't put much detail into the CHANGELOG, hope that's ok.

I have also renamed `master` to `main`.


buschbapti (approved):
Thanks for taking care of this and I think the README is good like this. We can now move to a proper CI way releasing more often as we do for `control-libraries`

domire8 (approved):
I trust you guys on this :wink:

PR #30, opened 2021-10-01
Update docker paradigm

domire8:
Same PR as in robot interface, now with one Dockerfile with stages and a build-server script

@eeberhard


buschbapti (approved):
Very good. Thanks a lot for that. I will have the same comment as in `robot_interface` I wonder whether the `ROS_VERSION` should appear in the tag or not.

PR #31, opened 2021-10-07
Refactor the shutdown logic

buschbapti:
This PR includes proper verification when shutting down the node both from lifecycle transition call and actual node destructor. It verifies in which state the node currently is and apply the transitions until it reaches the `FINALIZED` state.

Note that due to the implementation of the transitions in `ros2` it is actually not possible to use the `trigger_transition` function (shutting down transition has its proper logic from which you can't call deactivate and such; on destruction, the node can't publish its transition). Thus, we rely on our own `on_x` calls, for which we don't check the current state the node is.

This can be tested by simply loading one of the example and try combinations of `ctrl + c` / lifecycle calls.

### DIFF
--- a/source/modulo_core/src/Cell.cpp
+++ b/source/modulo_core/src/Cell.cpp
@@ -1,5 +1,7 @@
 #include "modulo_core/Cell.hpp"
 
+#include <lifecycle_msgs/msg/state.hpp>
+
 #include <state_representation/robot/JointPositions.hpp>
 #include <state_representation/robot/JointVelocities.hpp>
 #include <state_representation/exceptions/UnrecognizedParameterTypeException.hpp>
@@ -20,7 +22,11 @@ Cell::Cell(const rclcpp::NodeOptions& options) :
 }
 
 Cell::~Cell() {
-  this->on_shutdown();
+  RCUTILS_LOG_INFO_NAMED(get_name(), "Shutting down the node before destruction");
+  auto callback_code = this->on_shutdown(this->get_current_state());
+  if (callback_code == rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::FAILURE) {
+    RCUTILS_LOG_ERROR_NAMED(get_name(), "Error during the shutdown process, shutting down anyway.");
+  }
 }
 
 void Cell::reset() {
@@ -419,6 +425,28 @@ bool Cell::on_cleanup() {
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Cell::on_shutdown(const rclcpp_lifecycle::State& state) {
   RCUTILS_LOG_INFO_NAMED(get_name(), "on_shutdown() is called from state %s.", state.label().c_str());
+  uint8_t current_state = state.id();
+  // if the node is already shutdown just return success
+  if (state.id() == lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED) {
+    return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+  }
+  // check current state and eventually deactivate and deconfigure
+  if (current_state == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
+    RCUTILS_LOG_INFO_NAMED(get_name(), "Node is active, deactivating it before shutdown.");
+    auto callback_code = this->on_deactivate(this->get_current_state());
+    if (callback_code == rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::FAILURE) {
+      return callback_code;
+    }
+    current_state = lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE;
+  }
+  if (current_state == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE) {
+    RCUTILS_LOG_INFO_NAMED(get_name(), "Node is configured, deconfiguring it before shutdown.");
+    auto callback_code = this->on_cleanup(this->get_current_state());
+    if (callback_code == rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::FAILURE) {
+      return callback_code;
+    }
+    current_state = lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED;
+  }
   // call the proxy on_shutdown function
   if (!this->on_shutdown()) {
     RCLCPP_ERROR(get_logger(), "Shutdown failed.");
--- a/source/modulo_core/src/Cell.cpp
+++ b/source/modulo_core/src/Cell.cpp
@@ -9,6 +9,8 @@
 #include "modulo_core/utilities/utilities.hpp"
 #include "modulo_core/exceptions/UnconfiguredNodeException.hpp"
 
+using namespace rclcpp_lifecycle::node_interfaces;
+
 namespace modulo::core {
 
 Cell::Cell(const rclcpp::NodeOptions& options) :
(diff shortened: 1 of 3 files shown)


source/modulo_core/src/Cell.cpp
```diff
@@ -419,6 +425,28 @@ bool Cell::on_cleanup() {
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Cell::on_shutdown(const rclcpp_lifecycle::State& state) {
   RCUTILS_LOG_INFO_NAMED(get_name(), "on_shutdown() is called from state %s.", state.label().c_str());
+  uint8_t current_state = state.id();
```
domire8, line 428: I might be wrong here but it seems to me that the `lifecycle_msgs::msg::State` is an enum (with uint8 type). Wouldn't it be nicer to create the variable `current_state` as that enum instead of the underlying type?

eeberhard: lifecycle_msgs::msg::State is not an enum but rather a `static constexpr` that defines the symbol (e.g. `lifecycle_msgs::msg::State::PRIMARY_STATE_UNKNOWN` to have a value `uint8_t 0`

source/modulo_core/src/Cell.cpp
```diff
@@ -20,7 +22,11 @@ Cell::Cell(const rclcpp::NodeOptions& options) :
 }
 
 Cell::~Cell() {
-  this->on_shutdown();
+  RCUTILS_LOG_INFO_NAMED(get_name(), "Shutting down the node before destruction");
+  auto callback_code = this->on_shutdown(this->get_current_state());
+  if (callback_code == rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::FAILURE) {
+    RCUTILS_LOG_ERROR_NAMED(get_name(), "Error during the shutdown process, shutting down anyway.");
+  }
```
eeberhard, line 29: `callback_code` value is not used for much here, plus there is both a `FAILURE` and `ERROR` case (though one is deprecated). Better would be:

```suggestion
  if (this->on_shutdown(this->get_current_state()) != rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS) {
    RCUTILS_LOG_ERROR_NAMED(get_name(), "Error during the shutdown process, shutting down anyway.");
  }
```

source/modulo_core/src/Cell.cpp
```diff
@@ -419,6 +425,28 @@ bool Cell::on_cleanup() {
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Cell::on_shutdown(const rclcpp_lifecycle::State& state) {
   RCUTILS_LOG_INFO_NAMED(get_name(), "on_shutdown() is called from state %s.", state.label().c_str());
+  uint8_t current_state = state.id();
+  // if the node is already shutdown just return success
+  if (state.id() == lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED) {
```
eeberhard, line 430: for consistency, since you copy this value to `current_state`, I would suggest you also use it here
```suggestion
  if (current_state == lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED) {
```

source/modulo_core/src/Cell.cpp
domire8: okay nevermind then

source/modulo_core/src/Cell.cpp
```diff
@@ -419,6 +425,28 @@ bool Cell::on_cleanup() {
...
+  if (current_state == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
+    RCUTILS_LOG_INFO_NAMED(get_name(), "Node is active, deactivating it before shutdown.");
+    auto callback_code = this->on_deactivate(this->get_current_state());
+    if (callback_code == rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::FAILURE) {
+      return callback_code;
+    }
+    current_state = lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE;
+  }
+  if (current_state == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE) {
+    RCUTILS_LOG_INFO_NAMED(get_name(), "Node is configured, deconfiguring it before shutdown.");
```
eeberhard, line 443: Let's stick to ros2 terminology as part of this PR 

```suggestion
    RCUTILS_LOG_INFO_NAMED(get_name(), "Node is inactive, cleaning up before shutdown.");
```

source/modulo_core/src/Cell.cpp
```diff
@@ -419,6 +425,28 @@ bool Cell::on_cleanup() {
 
 rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn Cell::on_shutdown(const rclcpp_lifecycle::State& state) {
   RCUTILS_LOG_INFO_NAMED(get_name(), "on_shutdown() is called from state %s.", state.label().c_str());
```
eeberhard, line 427: maybe a `using namespace rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface` would be nice to reduce clutter for the `CallbackReturn`.

source/modulo_core/src/Cell.cpp
```diff
@@ -419,6 +425,28 @@ bool Cell::on_cleanup() {
...
+  if (state.id() == lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED) {
+    return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::SUCCESS;
+  }
+  // check current state and eventually deactivate and deconfigure
+  if (current_state == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE) {
+    RCUTILS_LOG_INFO_NAMED(get_name(), "Node is active, deactivating it before shutdown.");
+    auto callback_code = this->on_deactivate(this->get_current_state());
+    if (callback_code == rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::FAILURE) {
+      return callback_code;
+    }
```
eeberhard, line 439: I would again suggest comparing to "not success", and the syntax can be condensed.
```suggestion
    if (auto callback_return = this->on_deactivate(this->get_current_state()) != CallbackReturn::SUCCESS) {
      return callback_return;
    }
```

source/modulo_core/src/Cell.cpp
```diff
@@ -419,6 +425,28 @@ bool Cell::on_cleanup() {
...
+    current_state = lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE;
+  }
+  if (current_state == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE) {
+    RCUTILS_LOG_INFO_NAMED(get_name(), "Node is configured, deconfiguring it before shutdown.");
+    auto callback_code = this->on_cleanup(this->get_current_state());
+    if (callback_code == rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::FAILURE) {
+      return callback_code;
+    }
```
eeberhard, line 447:
```suggestion
    auto callback_code = ;
    if (auto callback_return = this->on_cleanup(this->get_current_state()) != CallbackReturn::SUCCESS) {
      return callback_return;
    }
```

source/modulo_core/src/Component.cpp
```diff
@@ -17,10 +17,6 @@ Component::Component(const rclcpp::NodeOptions& options) : Cell(options) {
   this->on_init();
 }
 
-Component::~Component() {
-  this->on_shutdown();
```
eeberhard, line 21: Thanks for removing this. It prevents multiple invocations of on_shutdown (which wasn't necessarily doing any harm since on_shutdown used to be empty, but just printing "on shutdown of the Cell class called" twice, which was always annoying)

eeberhard (commented):
Thanks for implementing this. It should make destruction much safer overall! According to the inverse law of code review, I have quite a few small comments for the small number of lines changed.

PR #32, opened 2021-10-07
Modify build script for production by default

buschbapti:
This PR adopts the paradigm where the default layer should be production not development. This way you can directly serve a production server or a development one.

One thing to note is that serving a development server, then a production one will generate an error as the two images do not have the same name. Thus, docker will complain that the port is already use.

### DIFF
--- a/build-server.sh
+++ b/build-server.sh
@@ -1,20 +1,23 @@
 #!/bin/bash
 ROS_VERSION=foxy
 
-BUILD_PROD=false
-SERVE_REMOTE=false
-
 IMAGE_NAME=epfl-lasa/modulo
+IMAGE_TAG=latest
+
 REMOTE_SSH_PORT=4440
+SERVE_REMOTE=false
 
 HELP_MESSAGE="Usage: build.sh [-p] [-r]
 Options:
-  -p, --production       Build the production ready image on top of
-                         of the development one.
+  -d, --development      Only target the development layer to prevent
+                         sources from being built or tested
+
   -r, --rebuild          Rebuild the image(s) using the docker
                          --no-cache option
+
   -v, --verbose          Use the verbose option during the building
                          process
+
   -s, --serve            Start the remove development server
 "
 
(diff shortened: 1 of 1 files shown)


build-server.sh
```diff
@@ -34,12 +37,8 @@ while [[ $# -gt 0 ]]; do
 done
 
 docker pull ghcr.io/aica-technology/ros2-control-libraries:"${ROS_VERSION}"
-DOCKER_BUILDKIT=1 docker build --target development -t "${IMAGE_NAME}/development:latest" "${BUILD_FLAGS[@]}" .
-
-if [ "${BUILD_PROD}" = true ]; then
-  DOCKER_BUILDKIT=1 docker build --target production -t "${IMAGE_NAME}:latest" "${BUILD_FLAGS[@]}" .
-fi
+DOCKER_BUILDKIT=1 docker build -t "${IMAGE_NAME}/${IMAGE_TAG}" "${BUILD_FLAGS[@]}" . || exit 1
```
eeberhard, line 40: Name and tag should be colon separated, same applies to line 43 for serving
```suggestion
DOCKER_BUILDKIT=1 docker build -t "${IMAGE_NAME}:${IMAGE_TAG}" "${BUILD_FLAGS[@]}" . || exit 1
```

build-server.sh
```diff
@@ -34,12 +37,8 @@ while [[ $# -gt 0 ]]; do
...
-DOCKER_BUILDKIT=1 docker build --target development -t "${IMAGE_NAME}/development:latest" "${BUILD_FLAGS[@]}" .
-
-if [ "${BUILD_PROD}" = true ]; then
-  DOCKER_BUILDKIT=1 docker build --target production -t "${IMAGE_NAME}:latest" "${BUILD_FLAGS[@]}" .
-fi
+DOCKER_BUILDKIT=1 docker build -t "${IMAGE_NAME}:${IMAGE_TAG}" "${BUILD_FLAGS[@]}" . || exit 1
 
 if [ "${SERVE_REMOTE}" = true ]; then
-  aica-docker server "${IMAGE_NAME}/development:latest" -u ros2 -p "${REMOTE_SSH_PORT}"
+  aica-docker server "${IMAGE_NAME}/${IMAGE_TAG}" -u ros2 -p "${REMOTE_SSH_PORT}"
```
domire8, line 43:
```suggestion
  aica-docker server "${IMAGE_NAME}:${IMAGE_TAG}" -u ros2 -p "${REMOTE_SSH_PORT}"
```

PR #34, opened 2021-11-23
Support int and int array parameters

domire8:
Review by commit here, I did three things
- Update to galactic
- Add int and int array support for parameters
- formatting

The int and int array was needed for my GMR component in dynamic components, that's why I had to make this change.

@eeberhard


buschbapti (approved):
This seems good to me. We will have to refactor all of those at some point

PR #36, opened 2022-01-12
Fix the examples due to breaking changes in DS

buschbapti:
This PR fixes the breaking changes introduced by the refactor of the DS in `control_libraries`. I also took the liberty of refactoring a bit to avoid the confusion with test as those are really only examples. I have also added an helicoidal that is a combination of a ring and a linear DS to the end as reducing the radius to 0 was creating weird behaviors.

A few things I have noticed that are most probably coming from the DS library. Most of it works fine except for the following:

* **Circular DS**: Default value of the limit cycle is empty. Not a problem per se but might be nicer to have a (0,0,0) center with radius = 1 as default. Could stay like this as well though. Could be nice to have also a `center` parameter that just changes the pose of the center as for now you have to go through the `limit_cycle` which is an `Ellipsoid`.

* **Linear DS**: In theory, the `attractor` name could be anything and it works, yet if I don't set it as the same name as the in state in the examples it creates jumps on rviz. I think the problem is that the ouptut velocity has the same name as the attractor and not the same name as the input state which it should have.

* **Joints DS**: Example not working at the moment I have to investigate. Might be that the rest of the code is not up to date and not a DS problem. The attractor, however, needs to be of type `JointState` and not `JointPositions` as opposed to the Cartesian one where a `CartesianPose` is fine.

* **Moving Reference Frame**: Example not working due to incompatible frames. Probably the rest of the code incorrect and not the DS. Have to investigate.


source/modulo_core/README.md
```diff
@@ -57,4 +57,4 @@ public:
 
 No need to do anything else. The step function is called periodically at a frequency defined in the node under the `period` argument. And the `current_state` variable will be published over `robot/joint_state` channel on ros2 network automatically.
 
-We recommand to keep this step function short, especially if you use a small period. Otherwise, latency will occur in your control loop. Complete examples of control loop including multiple nodes are written in the [tests](./tests/) for each representation spaces.
+We recommand to keep this step function short, especially if you use a small period. Otherwise, latency will occur in your control loop. Complete examples of control loop including multiple nodes are written in the [tests](examples/) for each representation spaces.
```
eeberhard, line 60:
```suggestion
We recommend keeping this step function short, especially if you use a small period. Otherwise, latency will occur in your control loop. Complete examples of control loops including multiple nodes are written in the [tests](examples/) for each representation spaces.
```

eeberhard (commented):
The migration to the factory DS pattern looks very good overall, and of course renaming "tests" to "examples" is good in my book. I have not tested this locally yet.

As long as the image itself builds successfully and the main examples are working, then it's good in my opinion. But, if you want to fix up the issues you mentioned in the summary then I'll wait for you to update it.

domire8 (dismissed):
This is fine for me too, I fixed the problem with the wrong name in the PointAttractor as well as the type of the JointState DS attractor in control libraries. 
I think it can go in like that so we can continue propagating the changes, the examples can be fixed in parallel.

However, as I said I don't think that the limit cycle of the circular DS should have a default value, but it would maybe be nice to have the option to set the center and the radiuses directly

PR #41, opened 2022-03-23
Modulo translators

domire8:
This is a first PR of many for the refactor of modulo. I created a branch `refactor/modulo` from develop for those PRs, to keep develop as is.

What I did here is the following:
- Create a new package (temporarily) named `modulo_new_core` and implement the `modulo_new_core::translators` functions to transform ROS messages to state rep types and vice versa - in both cpp and python.
- Use the `modulo_new_core::translators` in the original modulo package to prove that it's working.

The diff looks big but most of it comes from the `ReadStateConversion`  and `WriteStateConversion` that haven't been changed a lot


source/modulo_new_core/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+
+# find dependencies
+find_package(ament_cmake_auto REQUIRED)
+find_package(ament_cmake_python REQUIRED)
+
+ament_auto_find_build_dependencies()
+
+include_directories(include)
+
+# update below to change the component
```
eeberhard, line 26:
```suggestion
```

source/modulo_new_core/include/modulo_new_core/translators/ReadStateConversion.hpp
```diff
@@ -1,35 +1,24 @@
...
-#include <state_representation/space/joint/JointState.hpp>
 #include <state_representation/space/cartesian/CartesianPose.hpp>
 #include <state_representation/space/cartesian/CartesianState.hpp>
 #include <state_representation/space/cartesian/CartesianTwist.hpp>
 #include <state_representation/space/cartesian/CartesianWrench.hpp>
-#include <state_representation/space/dual_quaternion/DualQuaternionPose.hpp>
-#include <state_representation/space/dual_quaternion/DualQuaternionTwist.hpp>
-#include <state_representation/trajectories/Trajectory.hpp>
-
-#include "modulo_core/communication/EncodedState.hpp"
```
eeberhard, line 30: Would be nice to move this header along as part of this PR, so that we can continue to use the typedef for EncodedState in these files

source/modulo_new_core/modulo_new_core/translators/read_state_conversion.py
```diff
@@ -0,0 +1,73 @@
...
+        state.set_reference_frame(msg.header.frame_id)
+    elif isinstance(msg, geometry.WrenchStamped):
+        read_msg(state, msg.wrench)
+        state.set_reference_frame(msg.header.frame_id)
+    else:
+        raise RuntimeError("This message type is not supported")
+    return state
+
+
+def read_clproto_msg(msg: std_msgs.msg.UInt8MultiArray):
```
eeberhard, line 69: A `py` "header" for the EncodedState typedef would be nice as well for consistency

source/modulo_new_core/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+# find dependencies
+find_package(ament_cmake_auto REQUIRED)
+find_package(ament_cmake_python REQUIRED)
+
```
eeberhard, line 21: Shouldn't there be a `find_package(control_libraries ...)` and a `find_library(clproto)` somewhere here, or am I missing something?

source/modulo_new_core/src/translators/ReadStateConversion.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+void read_msg(state_representation::CartesianState& state, const geometry_msgs::msg::WrenchStamped& msg) {
+  state.set_reference_frame(msg.header.frame_id);
+  read_msg(state, msg.wrench);
+}
+
+void read_msg(state_representation::JointState& state, const sensor_msgs::msg::JointState& msg) {
+  state.set_names(msg.name);
+  if (!msg.position.empty()) state.set_positions(Eigen::VectorXd::Map(msg.position.data(), msg.position.size()));
```
eeberhard, line 72: Our style guide does not allow inline `if` statements 
```suggestion
  if (!msg.position.empty()) {
    state.set_positions(Eigen::VectorXd::Map(msg.position.data(), msg.position.size()));
  }
```

source/modulo_new_core/src/translators/WriteStateConversion.cpp
```diff
@@ -1,30 +1,65 @@
...
+
+static void write_quaternion(geometry_msgs::msg::Quaternion& msg, const Eigen::Quaterniond& quat) {
+  msg.w = quat.w();
+  msg.x = quat.x();
+  msg.y = quat.y();
+  msg.z = quat.z();
+}
+
+void write_msg(geometry_msgs::msg::Point& msg, const state_representation::CartesianState& state, const rclcpp::Time&) {
+  if (state.is_empty()) throw EmptyStateException(state.get_name() + " state is empty while attempting to publish it");
```
eeberhard, line 32: The inline `if` statements in this file should also be split with newlines and braces

source/modulo_new_core/src/translators/WriteStateConversion.cpp
```diff
@@ -1,30 +1,65 @@
...
+void write_msg(geometry_msgs::msg::Accel& msg, const state_representation::CartesianState& state, const rclcpp::Time&) {
+  if (state.is_empty()) throw EmptyStateException(state.get_name() + " state is empty while attempting to publish it");
+  write_vector3(msg.linear, state.get_linear_acceleration());
+  write_vector3(msg.angular, state.get_angular_acceleration());
+}
+
+void write_msg(geometry_msgs::msg::AccelStamped& msg, const state_representation::CartesianState& state, const rclcpp::Time& time) {
+  msg.header.stamp = time;
+  msg.header.frame_id = state.get_reference_frame();
+  write_msg(msg.accel, state, time);
```
eeberhard, line 55: You should invoke the other method first, since it checks for emptiness. Otherwise you only partially mutate the `msg`, for example with a potentially unexpected reference frame. Better to throw early without modifying the output argument.
```suggestion
  write_msg(msg.accel, state, time);
  msg.header.stamp = time;
  msg.header.frame_id = state.get_reference_frame();
```
I see it was already this way for the other stamped functions in this file, so you were just following the previous template. However I would encourage you to refactor the other functions to improve the behaviour in this manner.

domire8:
@eeberhard `EncodedState.hpp` is now in the new package too, but directly under `include/modulo_new_core` and not `include/modulo/communication`

source/modulo_new_core/test/cpp_tests/translators/test_translators.cpp
```diff
@@ -0,0 +1,175 @@
...
+}
+
+static void expect_quaternion_equal(const Eigen::Quaterniond& q1, const geometry_msgs::msg::Quaternion& q2) {
+  EXPECT_FLOAT_EQ(q1.x(), q2.x);
+  EXPECT_FLOAT_EQ(q1.y(), q2.y);
+  EXPECT_FLOAT_EQ(q1.z(), q2.z);
+  EXPECT_FLOAT_EQ(q1.w(), q2.w);
+}
+
+class TestAicaMsgs : public ::testing::Test {
```
eeberhard, line 29: I see you previously had a different name for this package 😄 By convention, test fixture classes should be named `FooTest`. I would suggest something along the lines of:
```suggestion
class TranslatorTest : public ::testing::Test {
```

source/modulo_new_core/test/cpp_tests/translators/test_translators.cpp
```diff
@@ -0,0 +1,175 @@
...
+  auto new_state = state_representation::JointState("test", {"1", "2", "3"});
+  read_msg(new_state, msg);
+  EXPECT_TRUE(new_state.get_positions().isApprox(joint_state_.get_positions()));
+  EXPECT_TRUE(new_state.get_velocities().isApprox(joint_state_.get_velocities()));
+  EXPECT_TRUE(new_state.get_torques().isApprox(joint_state_.get_torques()));
+  for (unsigned int i = 0; i < joint_state_.get_size(); ++i) {
+    EXPECT_EQ(new_state.get_names().at(i), joint_state_.get_names().at(i));
+  }
+}
+
```
eeberhard, line 166: would be good to have an `EXPECT_THROW` for empty state as a test case

eeberhard (commented):
Some comments but it looks quite good as the first step! Really useful to have the equivalent C++ and Python read and write msgs available.

source/modulo_new_core/test/cpp_tests/translators/test_translators.cpp
domire8: hehe you caught me

domire8:
Thanks for the review, I was able to improve quite a lot of things thanks to that.

I'm currently doing  `find_package(control_libraries 5.0.0 REQUIRED COMPONENTS state_representation)` which is not on main yet so I have to use the `galactic-devel` branch for the base image, not sure if thats a good idea or if its ok to assume we'll do a release on control libraries soon

domire8:
As you can see, I had to rebase due to #42 and I wasn't sure how I should do that but at the end I decided to do the following:
- As it was introduced in this PR, the read and write state conversion is deleted from `modulo_core` and can now be found in `modulo_new_core`
- As a consequence, `modulo_core` now depends on `modulo_new_core`, so whenever you want to build `modulo_core` you also need to build `modulo_new_core` 
- `modulo_components` will for now depend on both `modulo_core` and `modulo_new_core` until everything has been refactored to `modulo_new_core`.
- As seen in the most recent commit, we need now 3 different stages to develop, one with no packages built, an intermediate stage with `modulo_core` and hence `modulo_new_core` built, and the "production" stage with everything build. I therefore added a new option `-c|--core` for the intermediate stage to the build script in the most recent commit.

Now I know this is not very neat but we are still working on a refactor branch, and not on develop and we'll approach that big refactor from many different directions so I think it works like that for now. But I'm open to better solutions if they are not too time consuming :smile:

eeberhard (approved):
Cool, I think that's good to go in my opinion since it's on a refactor branch!

PR #42, opened 2022-03-24
Create modulo_components package

domire8:
Because I started working on the refactor I thought I would add the modulo_components package with those mostly empty files as early as possible such that they are available to everybody who wants to work on this refactor.

You can see that I created this package from the template component package so the python files and cpp tests contain code, but the package builds anyway because its commented out in the CMakeLists. What's most important are the cpp files for `ComponentInterface`, `LifecycleComponent` and `Component`


eeberhard (commented):
Good start to have the skeleton structure, and it's nice to define the headers and classes at the interface level first.

However I would avoid adding placeholder or unneeded code at this point. Keep the stubs and skeletons but nothing else. This means the commented cmakelists and component tests should be trimmed out. I would also say also either replace `py_component.py` with the appropriate class definitions for Python `Component` and `LifecycleComponent` (implementation can be absent), or remove that part entirely for now.

Dockerfile
```diff
@@ -1,10 +1,17 @@
...
+# upgrade ament_cmake_python
+RUN sudo apt update && sudo apt install -y ros-${ROS_DISTRO}-ament-cmake-python && sudo rm -rf /var/lib/apt/lists/*
+
 WORKDIR ${HOME}/ros2_ws
 # copy sources and build ROS workspace with user permissions
 WORKDIR ${HOME}/ros2_ws/
-COPY --chown=${USER} ./source/ ./src/
+COPY --chown=${USER} ./source/modulo_core ./src/modulo_core
+# build modulo
+RUN /bin/bash -c "source /opt/ros/$ROS_DISTRO/setup.bash; colcon build"
```
eeberhard, line 12: I assume you build the core first so that you can develop the components (with core dependencies) without having to compile the components package (which might be broken during development). However it's not ideal that the development layer builds one package and not the other.

For cleanness and flexibility, I would recommend you split this into multiple layers, something along the lines of:

```dockerfile
FROM ros2-control-libraries as dependencies
WORKDIR ${HOME}/ros2_ws
RUN sudo apt update && sudo apt install -y ros-${ROS_DISTRO}-ament-cmake-python && sudo rm -rf /var/lib/apt/lists/*

FROM dependencies as modulo-core
COPY --chown=${USER} ./source/modulo_core ./src/modulo_core
RUN /bin/bash -c "source /opt/ros/$ROS_DISTRO/setup.bash; colcon build --packages-select modulo_core"

FROM modulo-core as modulo-components
COPY --chown=${USER} ./source/modulo_components ./src/modulo_components
RUN /bin/bash -c "source /opt/ros/$ROS_DISTRO/setup.bash; colcon build --packages-select modulo_components"
```

eeberhard (approved):
Thanks for trimming this further. I think it's pretty much good to go as a starting point. I just have a minor suggestion on the dockerfile.

PR #43, opened 2022-04-01
Add MessagePairInterface and MessagePair including tests

domire8:
So this is what I have so far for the message interface and I was hoping to just get a quick feedback if this is going into the right direction or if I'm getting something wrong.  

Note that I don't include the encoded state (uint8_multi_arrray) for now and that I only focus on the things that are required to publish something, not for subscribing too.

Also I'm not happy with some of the function names so far, but I'd rather be verbose at the beginning and then we can cut stuff down once we have the whole framework.

@buschbapti

### DIFF
--- a/source/modulo_new_core/include/modulo_new_core/communication/MessagePair.hpp
+++ b/source/modulo_new_core/include/modulo_new_core/communication/MessagePair.hpp
@@ -1,0 +1,38 @@
+#pragma once
+
+#include "modulo_new_core/communication/MessagePairInterface.hpp"
+
+#include <std_msgs/msg/bool.hpp>
+#include <std_msgs/msg/float64.hpp>
+#include <std_msgs/msg/float64_multi_array.hpp>
+#include <std_msgs/msg/int64.hpp>
+#include <std_msgs/msg/string.hpp>
+#include <std_msgs/msg/u_int8_multi_array.hpp>
+
+namespace modulo_new_core::communication {
+
+template<typename MsgT, typename DataT>
+class MessagePair : public MessagePairInterface {
+private:
+  std::shared_ptr<DataT> data_;
+
+public:
+  explicit MessagePair(std::shared_ptr<DataT> data);
+
+  [[nodiscard]] MsgT get_message() const;
+
+  [[nodiscard]] std::shared_ptr<DataT> get_data() const;
+};
+
+template<typename MsgT, typename DataT>
+MsgT MessagePair<MsgT, DataT>::get_message() const {
+  auto msg = MsgT();
+  msg.data = *this->data_;
+  return msg;
+}
+
+template<typename MsgT, typename DataT>
+std::shared_ptr<DataT> MessagePair<MsgT, DataT>::get_data() const {
+  return this->data_;
+}
+}// namespace modulo_new_core::communication
--- a/source/modulo_new_core/include/modulo_new_core/communication/MessagePairInterface.hpp
+++ b/source/modulo_new_core/include/modulo_new_core/communication/MessagePairInterface.hpp
@@ -1,0 +1,58 @@
+#pragma once
+
+#include <memory>
+#include "modulo_new_core/communication/MessageType.hpp"
+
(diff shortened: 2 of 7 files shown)


source/modulo_new_core/include/modulo_new_core/communication/MessagePair.hpp
```diff
@@ -0,0 +1,37 @@
...
+public:
+  explicit MessagePair(std::shared_ptr<DataT> data);
+
+  [[nodiscard]] MsgT get_message() const;
+
+  [[nodiscard]] std::shared_ptr<DataT> get_data() const;
+};
+
+template<typename MsgT, typename DataT>
+MsgT MessagePair<MsgT, DataT>::get_message() const {
```
eeberhard, line 27: Ultimately, the `translators` namespace should have corresponding `write_msg(MsgT&, const DataT&)` templates or overloads that you simply call here as `write_msg(msg, *this->data_)`. This avoids any translation logic from being implemented at this layer. Maybe you can mark that with a TODO comment so it's clear.

source/modulo_new_core/include/modulo_new_core/communication/MessagePair.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+template<typename MsgT, typename DataT>
+class MessagePair : public MessagePairInterface {
+private:
+  std::shared_ptr<DataT> data_;
+
+public:
+  explicit MessagePair(std::shared_ptr<DataT> data);
+
+  [[nodiscard]] MsgT get_message() const;
```
eeberhard, line 21: Thinking ahead to the needs of a subscriber, the name of this method should maybe be adjusted.

`get_message` -> `write_message` (creates a message from the data, internally calls `translators::write_msg`)
`read_message(MsgT)` (updates the data from a message, internally calls `translators::read_msg`)

```suggestion
  [[nodiscard]] MsgT write_message() const;
```

source/modulo_new_core/include/modulo_new_core/communication/MessagePairInterface.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  virtual ~MessagePairInterface() = default;
+
+  MessagePairInterface(const MessagePairInterface& message_pair) = default;
+
+  template<typename MsgT, typename DataT>
+  std::shared_ptr<MessagePair<MsgT, DataT>> get_message_pair(bool validate_pointer = true);
+
+  template<typename MsgT, typename DataT>
+  MsgT get_message_pair_message();
```
eeberhard, line 24: For this method name I would also suggest `write_message()`.
```suggestion
  MsgT write_message();
```

source/modulo_new_core/include/modulo_new_core/communication/MessagePairInterface.hpp
```diff
@@ -0,0 +1,54 @@
...
+};
+
+template<typename MsgT, typename DataT>
+inline std::shared_ptr<MessagePair<MsgT, DataT>> MessagePairInterface::get_message_pair(bool validate_pointer) {
+  std::shared_ptr<MessagePair<MsgT, DataT>> message_pair_ptr;
+  try {
+    message_pair_ptr = std::dynamic_pointer_cast<MessagePair<MsgT, DataT>>(this->shared_from_this());
+  } catch (const std::exception& ex) {
+    if (validate_pointer) {
+      // TODO do we need a name for the MessagePairInterface?
```
eeberhard, line 39: I don't think so. Because it's essentially just a conversion wrapper around some referenced data, it should remain anonymous. This makes a meaningful error message difficult, but this is internal enough that we also don't expect this error to happen to normal users, so it doesn't need to be "friendly".

eeberhard (approved):
Looks quite good, and is definitely going in the right (or should I say _write_ 😉) direction. No required changes, just a few comments. Up to you if you want to keep the function names as they are for now, but I just left my suggestion.

Happy for you to merge once you address what you see fit.

PR #44, opened 2022-04-04
Use translators in MessagePair

domire8:
In order to use the translator functions in `MessagePair::write_message()`, I had to implement those translations first, which is what I'm doing here.
Additionally, I added two exceptions to handle pointer casting better.

A point that we can maybe discuss if if the MessagePair should hold it's own clock (initialized from the constructor), or if the `write_message` should take an argument which would be either time or clock

@buschbapti


eeberhard (approved):
Looks good to me. Instinctively I would tend to prefer the clock / time to be an argument rather than a class property, e.g. `MessagePair::write_message(const rclcpp::Time& time)`, which makes it a bit of a "lighter" container. The publisher wrapper would then hold a reference to a clock. But thinking about it more, given that these are all just shared references, there's no problem having it as a class property, and it has the benefit of making `write_message` much easier to call.

PR #45, opened 2022-04-05
Add EncodedState support

domire8:
Here I remove the specializations of the MessagePair constructor to allow that the `DataT` can be any of our state representation types that are supported by clproto. Hence I also add now the EncodedState to the enum of supported message types


eeberhard (approved):
Nice, looks good!

buschbapti (approved):

PR #47, opened 2022-04-07
Add publisher interface and handler

domire8:
So those are the files for publisher interface and handler. Review by commit. The first commit is just improving the message interface slightly.

To keep the PR light, I didn't include the implementation of the PublisherInterface yet, but I have it


source/modulo_new_core/include/modulo_new_core/communication/PublisherHandler.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+template<typename PubT, typename MsgT>
+class PublisherHandler : public PublisherInterface {
+public:
+  PublisherHandler(PublisherType type, std::shared_ptr<PubT> publisher);
+
+  void publish(const MsgT& message) const;
+
+private:
+  std::shared_ptr<PubT> publisher_;
```
eeberhard, line 15: At the PublisherHandler level, the `DataT` is still needed right? And when we `add_publisher` we use the data type to create both the handler and the message pair 
```suggestion
  std::shared_ptr<PubT<DataT>> publisher_;
```

eeberhard: sorry, I guess it should actually be `MsgT`

domire8: No, you cannot do <PubT<MsgT>>...

domire8: If you could to that, it would simplify a lot haha..maybe it gets clearer if I push the implementation and the tests?

eeberhard: Yeah I need to rethink some things at the top level. Here it makes sense that `PubT` refers to the specific publisher type (like `rclcpp::Publisher<std_msgs::msg::Bool>` or whatever), but the `PubT` we have in the top-level ComponentInterface was not supposed to work like that.

So for now I would suggest we remove `PubT` as a top-level template in `ComponentInterface` and just use the `PublisherType type` as a construction argument that the derived classes set

domire8:
Rebased and ready

source/modulo_components/include/modulo_components/ComponentInterface.h
```diff
@@ -82,6 +85,8 @@ class ComponentInterface : NodeT {
 
   void step();
 
+  modulo_new_core::communication::PublisherType publisher_type_;
```
eeberhard, line 88: this should be private and immutable by derived classes after instantiation

eeberhard (approved):
Thanks for that update. This looks good to me as the next step.

PR #48, opened 2022-04-07
Add tests for the predicate interface

buschbapti:
This PR simply provides the functions to add, get and set predicates, storing them in a map of `std::variant`. For now, I have removed the inheritance to `ParametersMap` that is otherwise annoying for testing and implementation.


buschbapti:
This has now been rebased and I changed the name of the PR to now reflect that this just add tests.

source/modulo_components/test/cpp/test_component_interface.cpp
```diff
@@ -0,0 +1,65 @@
+#include <gtest/gtest.h>
+#include <rclcpp/node.hpp>
+#include <rclcpp/publisher.hpp>
+#include <rclcpp/node_options.hpp>
+
+#include "modulo_components/ComponentInterface.h"
+#include "modulo_new_core/EncodedState.hpp"
+
+namespace modulo_components {
+class TestComponentInterface : public ::testing::Test {
```
eeberhard, line 10: Test fixture classes should follow the naming convention:
```suggestion
class ComponentInterfaceTest : public ::testing::Test {
```

source/modulo_components/include/modulo_components/ComponentInterface.h
```diff
@@ -19,19 +20,58 @@
 namespace modulo_components {
 
 template<class NodeT, typename PubT>
-class ComponentInterface : public state_representation::ParameterMap, public NodeT {
+class ComponentInterface : NodeT {
+  friend class TestComponentInterface;
+  FRIEND_TEST(TestComponentInterface, test_add_bool_predicate);
+  FRIEND_TEST(TestComponentInterface, test_add_function_predicate);
+  FRIEND_TEST(TestComponentInterface, test_get_predicate_value);
+  FRIEND_TEST(TestComponentInterface, test_set_predicate_value);
```
eeberhard, line 28: if the fixture class is declared as a friend, I would prefer you define the appropriate visibility in that class rather than declaring each test case as a friend. This otherwise won't scale well. 
```suggestion
  friend class ComponentInterfaceTest;
```

source/modulo_components/test/cpp/test_component_interface.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  void TearDown() override {
+    component_->predicates_.clear();
+  }
+
+  static ComponentInterface<rclcpp::Node, rclcpp::Publisher<modulo_new_core::EncodedState>>* component_;
+};
+
+ComponentInterface<rclcpp::Node, rclcpp::Publisher<modulo_new_core::EncodedState>>
+    * TestComponentInterface::component_ = nullptr;
```
eeberhard, line 28: Why do you use a static instance here and not just a regular property? It's a bit weird to read like this, and I would have thought this wouldn't be necessary for the friend behaviour.

Also, you shouldn't use raw pointer. Since we are internally using `std::shared_from_this` in the class, it's imperative that you actually hold a reference to the instance through a smart pointer, otherwise things can get weird (with no static or run-time assertion possible).

Finally, the intention was to pass a publisher typename symbol (e.g. Publisher or lifecycle::Publisher), rather than a class with a message type like `rclcpp::Publisher<modulo_new_core::EncodedState>`. So, overall, I feel like it should be:

```suggestion
  std::shared_ptr<ComponentInterface<rclcpp::Node, rclcpp::Publisher>> component_ = nullptr;
};
```

Let me know if I'm missing something

source/modulo_components/test/cpp/test_component_interface.cpp
```diff
@@ -0,0 +1,65 @@
...
+  auto value = std::get<bool>(predicate_iterator->second);
+  EXPECT_TRUE(value);
+}
+
+TEST_F(TestComponentInterface, test_add_function_predicate) {
+  component_->add_predicate("bar", [&]() { return false; });
+  auto predicate_iterator = component_->predicates_.find("bar");
+  EXPECT_TRUE(predicate_iterator != component_->predicates_.end());
+  auto value = std::get<std::function<bool(void)>>(predicate_iterator->second);
+  EXPECT_FALSE((value)());
```
eeberhard, line 43:
```suggestion
  auto value_callback = std::get<std::function<bool(void)>>(predicate_iterator->second);
  EXPECT_FALSE((value_callback)());
```

eeberhard (commented):
Really nice start. I have some suggestions, and I don't know if all them will work the way I wrote them, but I think they are important to address.

source/modulo_components/test/cpp/test_component_interface.cpp
buschbapti: Alright so the problem is that `rclcpp::Publisher` are actually not instantiable as they are template. There is a `rclcpp::GenericPublisher` that is not templated. I was actually using that one before switching to the one from Dominic as I thought it was raising an error and forgot to change back.

source/modulo_components/include/modulo_components/ComponentInterface.h
buschbapti: I followed the example from the link you sent. I will try to check if that is sufficient and can work like this. Indeed that would be better.

source/modulo_components/include/modulo_components/exceptions/ImmutablePredicateException.hpp
```diff
@@ -0,0 +1,12 @@
+#pragma once
```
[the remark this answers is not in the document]
buschbapti: Totally. I forgot to remove it thanks.

source/modulo_components/include/modulo_components/ComponentInterface.h
buschbapti: I made the test and as I thought this does not work. If you want to call a protected function in a test function, this test function needs to be explicitly added as `FRIEND_TEST`

source/modulo_components/test/cpp/test_component_interface.cpp
buschbapti: Regarding the static pointer I just followed the recommendation on the link you shared:
https://github.com/google/googletest/blob/main/docs/advanced.md#sharing-resources-between-tests-in-the-same-test-suite

I believe this will be the right approach when we use the node to test publishers and such. However, I changed for `std::unique_ptr` which works fine.

source/modulo_components/test/cpp/test_component_interface.cpp
```diff
@@ -0,0 +1,65 @@
...
+    if (component_ == nullptr) {
+      component_ =
+          new ComponentInterface<rclcpp::Node, rclcpp::Publisher<modulo_new_core::EncodedState>>(rclcpp::NodeOptions());
+    }
+  }
+
+  void TearDown() override {
+    component_->predicates_.clear();
+  }
+
```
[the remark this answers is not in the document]
buschbapti: As each test cases need to be friend anyway, it is actually faster to just use the direct access to the private variable.

source/modulo_components/include/modulo_components/ComponentInterface.h
eeberhard: It would definitely work fine as long as the `ComponentInterfaceTest` fixture class provides the right accessors. Better for the test code to scale with each test requirement than the class header. I tested it locally to make sure that something like the following works if you only declare `friend class ComponentInterfaceTest;` and nothing else.

It also raises the question if these functions, which are designed to be a sort of user-interface for the developer, really need to be `protected` in the first place. It's rare that any user will want access to an instance of the component at run-time, and certainly is not possible from within our application at run-time. 

Anyway TL;DR I think expanding a wrapper class in the test suite is preferable to expanding friends for test cases in the public header.

```c++
#include ...

namespace modulo_components {
class ComponentInterfaceTest : public ::testing::Test {
protected:
  static void SetUpTestSuite() {
    rclcpp::init(0, nullptr);
  }

  void SetUp() override {
    component_ = std::make_shared<ComponentInterface<rclcpp::Node, rclcpp::Publisher<modulo_new_core::EncodedState>>>(rclcpp::NodeOptions());
  }

  void TearDown() override {
    component_->predicates_.clear();
  }

  std::map<std::string, utilities::PredicateVariant> get_predicate_map() {
    return component_->predicates_;
  }

  void add_predicate(const std::string& name, bool value) {
    component_->add_predicate(name, value);
  }

  void add_predicate(const std::string& predicate_name, const std::function<bool(void)>& predicate_function) {
    component_->add_predicate(predicate_name, predicate_function);
  }

  [[nodiscard]] bool get_predicate(const std::string& predicate_name) const {
    return component_->get_predicate(predicate_name);
  }

  /**
   * @brief Set the value of the predicate given as parameter
   * @param predicate_name the name of the predicate to retrieve from the
   * map of predicates
   * @param predicate_value the new value of the predicate
   */
  void set_predicate(const std::string& predicate_name, bool predicate_value) {
    component_->set_predicate(predicate_name, predicate_value);
  }

  std::shared_ptr<ComponentInterface<rclcpp::Node, rclcpp::Publisher<modulo_new_core::EncodedState>>> component_;
};

TEST_F(ComponentInterfaceTest, test_add_bool_predicate) {
  add_predicate("foo", true);
  auto predicates = get_predicate_map();
  auto predicate_iterator = predicates.find("foo");
  EXPECT_TRUE(predicate_iterator != predicates.end());
  auto value = std::get<bool>(predicate_iterator->second);
  EXPECT_TRUE(value);
}

TEST_F(ComponentInterfaceTest, test_add_function_predicate) {
  add_predicate("bar", [&]() { return false; });
  auto predicates = get_predicate_map();
  auto predicate_iterator = predicates.find("bar");
  EXPECT_TRUE(predicate_iterator != predicates.end());
  auto value_callback = std::get<std::function<bool(void)>>(predicate_iterator->second);
  EXPECT_FALSE((value_callback)());
}

TEST_F(ComponentInterfaceTest, test_get_predicate_value) {
  add_predicate("foo", true);
  EXPECT_TRUE(get_predicate("foo"));
  add_predicate("bar", [&]() { return false; });
  EXPECT_FALSE(get_predicate("bar"));
  // TODO suppress the no discard warning
  EXPECT_THROW(get_predicate("test"), exceptions::PredicateNotFoundException);
}

TEST_F(ComponentInterfaceTest, test_set_predicate_value) {
  add_predicate("foo", true);
  set_predicate("foo", false);
  EXPECT_FALSE(get_predicate("foo"));
  EXPECT_THROW(set_predicate("bar", true), exceptions::PredicateNotFoundException);
  add_predicate("bar", [&]() { return false; });
  set_predicate("bar", true);
  EXPECT_TRUE(get_predicate("bar"));
}
}
```

buschbapti: Right that is actually much better. I was a bit slow on the understanding there.

source/modulo_components/include/modulo_components/ComponentInterface.h
```diff
@@ -61,21 +96,22 @@ class ComponentInterface : public state_representation::ParameterMap, public Nod
 
 template<class NodeT, class PubT>
 ComponentInterface<NodeT, PubT>::ComponentInterface(const rclcpp::NodeOptions& options) :
-    rclcpp::Node(utilities::parse_node_name(options, "ComponentInterface")) {
-  this->declare_parameter("period", 1.0);
+    rclcpp::Node(utilities::parse_node_name(options, "ComponentInterface"), options) {
+  this->declare_parameter("period", 2.0);
```
buschbapti, line 100:
```suggestion
  this->declare_parameter("period", 1.0);
```

source/modulo_components/include/modulo_components/ComponentInterface.h
```diff
@@ -125,18 +156,34 @@ ComponentInterface<NodeT, PubT>::add_predicate(const std::string& name, const ut
...
+    throw exceptions::PredicateNotFoundException(predicate_name);
+  }
+  // try to get the value from the variant as a bool
+  auto* ptr_value = std::get_if<bool>(&predicate_iterator->second);
+  if (ptr_value) {
+    return *ptr_value;
+  }
+  // if previous check failed, it means the variant is actually a callback function
+  auto callback_function = std::get<std::function<bool(void)>>(predicate_iterator->second);
+  return (callback_function)();
```
domire8, line 182: Can we have something like this (proposed by Enrico in a previous PR):
```
if(const bool* val = std::get_if<bool>(&predicate.second)) {
  return *val;
} else if(const std::function<bool(void)>* val = std::get_if<std::function<bool(void)>>(&predicate.second)) {
  return *val();
}
```

domire8 (approved):
Looks good!

source/modulo_components/test/cpp/test_component_interface.cpp
```diff
@@ -7,59 +7,73 @@
...
-    * TestComponentInterface::component_ = nullptr;
+  bool get_predicate(const std::string& predicate_name) const {
+    return component_->get_predicate(predicate_name);
+  }
+
+  void set_predicate(const std::string& predicate_name, bool predicate_value) {
+    component_->set_predicate(predicate_name, predicate_value);
+  }
+
+  std::unique_ptr<ComponentInterface<rclcpp::Node, rclcpp::GenericPublisher>> component_;
```
eeberhard, line 40: Because we inherit from `std::enable_shared_from_this`, it really is necessary for the component interface to be held by a shared pointer and not a unique pointer. Otherwise, we'll have `std::bad_weak_ptr` or undefined exceptions at a later point. Unique pointers should not be used to manage nodes.

eeberhard (approved):
Thanks for the changes! This is a great extension with the tests.

PR #49, opened 2022-04-08
Fix modulo_core after StateType refactor

domire8:
Quick fix so we can keep modulo_core alive on the refactor branch

The base image tag in the build script should go back to `galactic-devel` once merged and rebuilt


eeberhard (approved):
Nice, thank you, that's better than just suppressing the package entirely

PR #50, opened 2022-04-08
Update predicate logic in ComponentInterface

buschbapti:
While working on the python implementation, it appeared that the predicate logic was slightly flawed and needed updating. In particular:

* `get_predicate` does not throw exceptions but return False in case of errors
* `set_predicate` does not do anything if the predicate does not exists already

Test cases have been updated accordingly.


domire8:
Let's close this soon, been open a long time. I can help rebasing if you want @buschbapti

buschbapti:
Thanks I believe I was waiting for @eeberhard but forgot to remind. I have rebased. However, @domire8 can you double check? There are new functionalities you have introduced since then that might impact it and require changers to be made.

domire8:
The rebase doesn't look good yet, mainly because of the two commits (PR 42 and 49) that are still in here

buschbapti:
Right I guess they can just be dropped actually

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -195,6 +194,20 @@ ComponentInterface<NodeT>::ComponentInterface(
...
+  this->has_tf_listener_ = NodeT::get_parameter("has_tf_listener");
+  this->has_tf_broadcaster_ = NodeT::get_parameter("has_tf_broadcaster");
+
+  if (has_tf_listener_.as_bool()) {
+    this->tf_buffer_ = std::make_shared<tf2_ros::Buffer>(this->get_clock());
+    this->tf_listener_ = std::make_shared<tf2_ros::TransformListener>(*this->tf_buffer_);
+  }
+  if (has_tf_broadcaster_.as_bool()) {
+    this->tf_broadcaster_ = std::make_shared<tf2_ros::TransformBroadcaster>(this->shared_from_this());
+  }
```
domire8, line 210: All this should be removed, this was not part of your original PR

buschbapti: It might have been actually but should not be here anyway. This should be added later in another PR and updated with the new parameters accessors.

domire8: No this was removed on purpose

buschbapti: ok

buschbapti: alright yes that makes a lot of sense

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -121,7 +120,7 @@ class ComponentInterface : NodeT {
    * map of predicates
    * @return the value of the predicate as a boolean
    */
-  [[nodiscard]] bool get_predicate(const std::string& predicate_name) const;
+  [[nodiscard]] bool get_predicate(const std::string& predicate_name);
```
eeberhard, line 123: is there a reason this can't be const anymore? I guess because of the clock for the throttled logger? It's fine in that case

eeberhard (approved):
Seems good, thanks @domire8 for catching the rebase issues

source/modulo_components/include/modulo_components/ComponentInterface.hpp
buschbapti: yes exactly. This is actually annoying. I wonder if ROS found an alternative to that.

eeberhard: I guess maybe you could instantiate a static clock inside the function. If it's scoped there and not in the class, the class could remain const. But it's not worth worrying about here for the time being

PR #51, opened 2022-04-08
Add the base ComponentInterface class in python

buschbapti:
This PR introduces the `ComponentInterface` class in python with testing of the predicate interface.

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -156,7 +156,6 @@ void ComponentInterface<NodeT>::add_variant_predicate(
     this->predicates_.at(name) = predicate;
   } else {
     this->predicates_.insert(std::make_pair(name, predicate));
-    // TODO add publisher with message interface
     this->predicate_publishers_.insert(
         std::make_pair(
             name, this->template create_publisher<std_msgs::msg::Bool>(this->generate_predicate_topic(name), 10)));
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -48,8 +48,8 @@ protected:
   void add_predicate(const std::string& predicate_name, const std::function<bool(void)>& predicate_function);
 
   /**
-   * @brief Get the value of the predicate given as parameter, if the predicate is not found or the callable function
-   * fails return false
+   * @brief Get the logical value of a predicate.
+   * @details If the predicate is not found or the callable function fails, the return value is false.
    * @param predicate_name the name of the predicate to retrieve from the
    * map of predicates
    * @return the value of the predicate as a boolean
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -48,7 +48,8 @@ protected:
   void add_predicate(const std::string& predicate_name, const std::function<bool(void)>& predicate_function);
 
   /**
-   * @brief Get the value of the predicate given as parameter
+   * @brief Get the value of the predicate given as parameter, if the predicate is not found or the callable function
+   * fails return false
    * @param predicate_name the name of the predicate to retrieve from the
    * map of predicates
    * @return the value of the predicate as a boolean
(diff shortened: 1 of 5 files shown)


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -0,0 +1,121 @@
...
+                                   description="If true, add a tf listener to enable calls to lookup transforms."))
+        self.declare_parameter('has_tf_broadcaster', False,
+                               ParameterDescriptor(
+                                   description="If true, add a tf broadcaster to enable the publishing of transforms."))
+
+        self._period = self.get_parameter('period').get_parameter_value().double_value
+        self._has_tf_listener = self.get_parameter('has_tf_listener').get_parameter_value().double_value
+        self._has_tf_broadcaster = self.get_parameter('has_tf_broadcaster').get_parameter_value().double_value
+        self._predicates = {}
+        self._predicates_publishers = {}
```
eeberhard, line 49: I would prefer to keep the general rule of "single, plural" for collections rather than "plural, plural". The "pencil sharpener" vs. "pencils sharpeners" example comes to mind.
```suggestion
        self._predicate_publishers = {}
```

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -48,26 +48,27 @@ class ComponentInterface : NodeT {
   void add_predicate(const std::string& predicate_name, const std::function<bool(void)>& predicate_function);
 
   /**
-   * @brief Get the value of the predicate given as parameter
+   * @brief Get the value of the predicate given as parameter, if the predicate is not found or the callable function
+   * fails return false
```
eeberhard, line 52: Avoid run-on sentences - if the brief description is longer than a sentence, use `@details`. I also would avoid "the predicate given as a parameter" in favour of something like "a / the named predicate", or even just "Get the logical value of a predicate". It's not necessary to specify "the passed parameter" for single-argument methods.

This would apply to the other docstrings here and in Python as well.

```suggestion
   * @brief Get the logical value of a predicate.
   * @details If the predicate is not found or the callable function fails, the return value is false.
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -0,0 +1,120 @@
...
+            self.get_logger().debug(f"Predicate {predicate_name} already exists, overwriting.")
+        else:
+            self._predicates_publishers[predicate_name] = self.create_publisher(Bool, self.__generate_predicate_topic(
+                predicate_name), 10)
+        self._predicates[predicate_name] = predicate_value
+
+    def get_predicate(self, predicate_name: str) -> bool:
+        """
+        Get the value of the predicate given as parameter, if the predicate is not found or the callable function fails
+        return false.
```
eeberhard, line 89:
```suggestion
        Get the value of the predicate given as parameter. If the predicate is not found or the callable function fails,
        return false.
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -0,0 +1,120 @@
...
+
+        :param predicate_name: the name of the predicate to retrieve from the map of predicates
+        :return: the value of the predicate as a boolean
+        """
+        if predicate_name not in self._predicates.keys():
+            self.get_logger().error(f"Predicate {predicate_name} does not exists, returning false.",
```
eeberhard, line 95:
```suggestion
            self.get_logger().error(f"Predicate {predicate_name} does not exist, returning false.",
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def set_predicate(self, predicate_name: str, predicate_value: Union[bool, Callable]):
+        """
+        Set the value of the predicate given as parameter, if the predicate is not found does not do anything.
+
+        :param predicate_name: the name of the predicate to retrieve from the map of predicates
+        :param predicate_value: the new value of the predicate as a bool or a callable function
+        """
+        if predicate_name not in self._predicates.keys():
+            self.get_logger().error(f"Predicate {predicate_name} does not exists, can't set it with a new value.",
```
eeberhard, line 117:
```suggestion
            self.get_logger().error(f"Cannot set predicate {predicate_name} with a new value because it does not exist.",
```

eeberhard (approved):
I only have some suggestions for the documentation, but overall this looks great! Nice work

source/modulo_components/modulo_components/component_interface.py
buschbapti: This has been updated in all occurences

domire8 (approved):

PR #52, opened 2022-04-11
Specialize all available MessagePairs and add read capability

domire8:
Here I specialize all available MessagePairs. As a consequence, I can remove the `MessageType` argument from the constructor. Additionally, I already add the `read_message` / `read` methods for the future subscription handler.

Note that the specializations for the `EncodedState` read/write are marked as TODOs because I'm waiting for other PRs to resolve those.

### DIFF
--- a/source/modulo_new_core/CMakeLists.txt
+++ b/source/modulo_new_core/CMakeLists.txt
@@ -27,6 +27,7 @@ find_library(clproto REQUIRED)
 include_directories(include)
 
 ament_auto_add_library(modulo_new_core SHARED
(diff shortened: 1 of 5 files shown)


eeberhard (approved):
I checked this yesterday, and I think it's good as incremental progress if you want to do this now and then resolve the translators at a later point.

domire8:
> I checked this yesterday, and I think it's good as incremental progress if you want to do this now and then resolve the translators at a later point.

Yes will do once we closed a few more PRs on control libraries

PR #53, opened 2022-04-12
Implementation of publisher interface

domire8:
Another small increment towards our new components.
This PR adds the implementation of the previously defined `publish` method for the `PublisherInterface`. Due to the fact that this method doesn't have arguments or templates, two switch cases are needed to get and publish the correct message with the correct publisher type.

### DIFF
--- a/source/modulo_new_core/include/modulo_new_core/communication/PublisherInterface.hpp
+++ b/source/modulo_new_core/include/modulo_new_core/communication/PublisherInterface.hpp
@@ -56,9 +56,4 @@ inline std::shared_ptr<PublisherHandler<PubT, MsgT>> PublisherInterface::get_pub
   return publisher_ptr;
 }
 
-static std::shared_ptr<PublisherInterface> create_publisher_interface(const std::shared_ptr<PublisherInterface>& publisher, const std::shared_ptr<MessagePairInterface>& message_pair) {
-  publisher->set_message_pair(message_pair);
-  return publisher;
-}
-
 }// namespace modulo_new_core::communication
--- a/source/modulo_new_core/include/modulo_new_core/communication/PublisherInterface.hpp
+++ b/source/modulo_new_core/include/modulo_new_core/communication/PublisherInterface.hpp
@@ -38,4 +38,28 @@ private:
   std::shared_ptr<MessagePairInterface> message_pair_;
 };
 
+template<typename PubT, typename MsgT>
+inline std::shared_ptr<PublisherHandler<PubT, MsgT>> PublisherInterface::get_publisher(bool validate_pointer) {
+  std::shared_ptr<PublisherHandler<PubT, MsgT>> publisher_ptr;
+  try {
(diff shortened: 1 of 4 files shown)


source/modulo_new_core/include/modulo_new_core/communication/PublisherInterface.hpp
```diff
@@ -38,4 +38,28 @@ class PublisherInterface : public std::enable_shared_from_this<PublisherInterfac
...
+    throw exceptions::InvalidPointerCastException(
+        "Unable to cast publisher interface to a publisher pointer of requested type"
+    );
+  }
+  return publisher_ptr;
+}
+
+static std::shared_ptr<PublisherInterface> create_publisher_interface(const std::shared_ptr<PublisherInterface>& publisher, const std::shared_ptr<MessagePairInterface>& message_pair) {
+  auto pub = std::shared_ptr<PublisherInterface>(publisher);
+  pub->set_message_pair(message_pair);
```
eeberhard, line 61: These lines are a bit weird. If you want to make a copy of the argument pointer to reference the same object, it would suffice to write `auto pub = publisher;` I understand the intention is to pass a pointer to a PublisherHandler as the arg, but since the argument type is declared in the function signature it doesn't really need to be duplicated. 

But, moreover, the `publisher` pointer that is passed to the function will be copy constructed anyway if a pointer to a derived type pointer (e.g. PublisherHandler) is passed, so another copy is not really useful. Finally keep in mind that a const pointer is not the same as a pointer to const. All the pointer copies still refer to the non-const object which can be modified through the pointer. 

So, you can just do:
```suggestion
  publisher->set_message_pair(message_pair);
```

source/modulo_new_core/test/cpp_tests/communication/test_publisher_handler.cpp
```diff
@@ -0,0 +1,77 @@
...
+  EXPECT_NO_THROW(publisher_interface->publish());
+
+  publisher_interface = create_publisher_interface(publisher_handler, msg_pair);
+  EXPECT_NO_THROW(publisher_interface->publish());
+}
+
+class PublisherTest : public ::testing::Test {
+protected:
+  void SetUp() {
+    rclcpp::init(0, nullptr);
```
eeberhard, line 35: For ROS context init and shutdown it may be preferable to use `static void SetUpTestSuite()` and `static void TearDownTestSuite()`. Test objects like the Node should still be constructed on each test case with `SetUp` for same initial conditions.

But, you only have one test case for now so it doesn't really matter.

eeberhard (approved):
Looks quite good! Only a few minor suggestions for code improvement.

source/modulo_new_core/include/modulo_new_core/communication/PublisherInterface.hpp
domire8: Hm yeah that makes sense, thanks

PR #57, opened 2022-04-18
Parameter translation improvements

eeberhard:
I eventually got around to finalizing the parameter translator with the Component requirements in mind. 

In the case of some ROS parameter types, such as double array, is it necessary to have a "reference" parameter to check if the translation should be a double vector, eigen vector or matrix. However, we sometimes want to avoid actually updating the value of this reference directly, for example, if we need to validate the value before setting. On the component side, I want the validation function to be using ParameterInterface only, so translation has to happen first. Because of this, I added a const reader that does not change the reference, and subsequently a value copy function that can be used to update the value of the reference post validation.

Hopefully the tests and the related component PR make the behaviour clear!

@buschbapti @domire8


source/modulo_new_core/src/translators/parameter_translators.cpp
```diff
@@ -106,120 +179,69 @@ std::shared_ptr<state_representation::ParameterInterface> read_parameter(const r
...
       break;
   }
   throw InvalidParameterException("Parameter " + parameter.get_name() + " could not be read!");
 }
 
-void read_parameter(
-    const rclcpp::Parameter& ros_parameter, std::shared_ptr<state_representation::ParameterInterface>& parameter
+std::shared_ptr<state_representation::ParameterInterface> read_parameter_const(
+    const rclcpp::Parameter& ros_parameter,
+    const std::shared_ptr<const state_representation::ParameterInterface>& parameter
```
domire8, line 189: What part here would go wrong in this function if the second argument wasn't shared pointer of `const` ParameterInterface?

domire8: Or would everything still work but we just want to be super explicit to be sure that the pointer is kept as the 'reference pointer' like you mentioned above?

eeberhard: Nothing would go wrong, but by allowing calls to non-const methods it adds a risk that the value might be modified accidentally. This is just making everything more explicit.

eeberhard:
This has been rebased on the new control libraries image and now all methods are working as desired (including matrix translation)

domire8 (approved):
Great, that works for me

PR #58, opened 2022-04-18
Component Parameters

eeberhard:
* Implement add_ and get_ methods for component parameters, using
the modulo parameter translators.

* Link clproto in CMakeLists because of parameter translators

* Add tests for component parameter behaviour

This PR follows #57 and applies the translation logic in the ComponentInterface.

@buschbapti @domire8

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -63,7 +63,7 @@ protected:
    * @param name The name of the parameter
    * @return The ParameterInterface pointer to a Parameter instance
    */
-  std::shared_ptr<state_representation::ParameterInterface> get_parameter(const std::string& name);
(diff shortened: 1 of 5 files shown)


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -33,6 +39,53 @@ class ComponentInterface : NodeT {
...
+   */
+  template<typename T>
+  void add_parameter(const std::string& name, const T& value);
+
+  /**
+   * @brief Get a parameter by name.
+   * @param name The name of the parameter
+   * @return The ParameterInterface pointer to a Parameter instance
+   */
+  std::shared_ptr<state_representation::ParameterInterface> get_parameter(const std::string& name);
```
domire8, line 66: Shouldn't those getters now be marked `const` as well or is this a different situation now?

domire8 (approved):

PR #60, opened 2022-04-19
Support write/read for MessagePair<EncodedState, State>

domire8:
Now that we have the most recent changes from control libraries (encode/decode shared pointer of State), we can use that in the translators and message pair to read and write the encoded state pair (and also publish it with the PublisherInteface)

@buschbapti


eeberhard (approved):
Seems good to me!

PR #62, opened 2022-04-22
Add methods to add output/publishers

domire8:
Lot of added lines due to template specifications, additional tests and switch cases. I'm adding here the methods to add the publishers to the components, given a signal name and data pointer. It's not entirely finished and cleaned up but for me the right place to get a first review. Review by commit.


eeberhard (commented):
Looks very good! I like how you are creating the topic parameter by default and using the `read_only` flag for fixed names. That makes things very consistent and simple.

I have a few minor comments but I think this a great step forward

domire8:
Thanks for the feedback @eeberhard, I hope I caught the exceptions as you were suggesting it.

What are we thinking about the parsing of the signal name? What is the expected input there?
```
  this->outputs_.insert(
      std::make_pair(
          signal_name, std::make_shared<PublisherInterface>(this->publisher_type_, message_pair)));
  this->add_parameter(
      signal_name + "_topic", "~/" + signal_name, "Output topic name of signal '" + signal_name + "'", fixed_topic
  );
```
Should we expect something like `desired_topic_name` (without slashes) or `/this_topic` (with one or possibly more slashes)? The way its implemented now would expect that there are no slashes in the `signal_name` argument but I think we have to support both options.

source/modulo_new_core/include/modulo_new_core/communication/PublisherHandler.hpp
```diff
@@ -22,8 +27,31 @@ template<typename PubT, typename MsgT>
 PublisherHandler<PubT, MsgT>::PublisherHandler(PublisherType type, std::shared_ptr<PubT> publisher) :
     PublisherInterface(type), publisher_(std::move(publisher)) {}
 
+template<typename PubT, typename MsgT>
+void PublisherHandler<PubT, MsgT>::on_activate() {
+  if (this->get_type() != PublisherType::LIFECYCLE_PUBLISHER) {
```
domire8, line 32: Is it okay if this throws or should it just be ignored?

source/modulo_components/src/LifecycleComponent.cpp
```diff
@@ -7,74 +7,99 @@ namespace modulo_components {
...
+  bool success = true;
+  for (auto const& [name, interface]: this->outputs_) {
+    try {
+      interface->deactivate();
+    } catch (const std::exception& ex) {
+      success = false;
+      RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to deactivate output '" << name << "': " << ex.what());
+    }
   }
+  return success;
```
eeberhard, line 102: I think you can avoid the intermediate variable in these functions and just return, since it's happening at the end of the function.
```suggestion
      RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to deactivate output '" << name << "': " << ex.what());
      return false;
    }
  }
  return true;
```

source/modulo_new_core/include/modulo_new_core/communication/PublisherHandler.hpp
eeberhard: to be honest I think it should just warn and return

eeberhard (commented):
>What are we thinking about the parsing of the signal name?

I agree the signal name should be validated, but keep in mind that the signal name is just a label and is different from the topic name/namespace. So, I think validating that it is lower_snake_case, with only alphanumeric and _ characters is good. I think appending `_topic` to the parameter is very appropriate. 

So we would have, for example, a `wrench_command` signal with a `wrench_command_topic` parameter, and a `cartesian_state` signal with a `cartesian_state_topic` parameter. The default topic names are `~/wrench_command` and `~/cartesian_state`. I don't particularly want parameter names to contain `/`.

Now there might be rare cases where the default topic name should be different from the automatically generated one (for example, some subscriber to a global channel like `/tf`). To support these cases, I guess you can have a fourth optional string argument for `create_output(..., std::string default_topic = "")`. If it's empty, use `"~/" + signal_name`, and otherwise use `default_topic`.

The rest is looking quite good

source/modulo_new_core/include/modulo_new_core/communication/PublisherHandler.hpp
domire8: We can't warn here though, except from using `std::cout`

source/modulo_components/src/LifecycleComponent.cpp
domire8: But it's in for loops so I was thinking we should try to activate/deactivate as many outputs as possible and not return as soon as one output fails. Not good?

eeberhard: Oh yes of course, I missed the alignment of the braces because of the diff formatting, indeed that's better like you have it

source/modulo_components/include/modulo_components/utilities/utilities.hpp
```diff
@@ -25,6 +25,10 @@ static std::string parse_string_argument(const std::vector<std::string>& args, c
   return result;
 }
 
+static std::string parse_signal_name(const std::string& signal_name) {
+  return signal_name;
```
eeberhard, line 29: Nice function haha. Maybe add a //TODO here 😁

eeberhard (approved):
Looks good. I think it can be merged before getting any bigger.

domire8:
Ok yes we should merge now, I think I addressed everyting

eeberhard (approved):
Thanks for those last changes!

PR #64, opened 2022-04-27
Subscription handler/interface and add_input

domire8:
Quite a big step IMO, I added a SubscriptionInterface and SubscriptionHandler to `modulo_new_core` in order to be able to create and store ROS subscriptions in non-templated pointers.
Subscriptions can be added in two ways, see `add_input` in ComponentInterface:
- Either by providing a shared pointer to a variable, which creates a message pair to read the state in the subscription callback
- or by providing a callback function

I would like to update the tests so they run on both the LifecycleNode and the Node but that can be done in a subsequent PR

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -30,6 +30,7 @@ template<class NodeT>
 class ComponentInterface : public NodeT {
 public:
   friend class ComponentInterfacePublicInterface;
+  friend class ComponentInterfaceParameterPublicInterface;
 
   /**
    * @brief Constructor from node options
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -139,12 +139,28 @@ protected:
    */
   void set_predicate(const std::string& predicate_name, const std::function<bool(void)>& predicate_function);
 
+  /**
+   * @brief Add and configure an input signal of the component.
+   * @tparam DataT Type of the data pointer
+   * @param signal_name Name of the output signal
+   * @param data Data to transmit on the output signal
+   * @param fixed_topic If true, the topic name of the output signal is fixed
+   * @param default_topic If set, the default value for the topic name to use
+   */
   template<typename DataT>
   void add_input(
       const std::string& signal_name, const std::shared_ptr<DataT>& data, bool fixed_topic = false,
       const std::string& default_topic = ""
   );
 
+  /**
+   * @brief Add and configure an input signal of the component.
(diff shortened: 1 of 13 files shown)


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -425,6 +455,98 @@ void ComponentInterface<NodeT>::set_predicate(
...
+      }
+      case MessageType::ENCODED_STATE: {
+        auto subscription_handler = std::make_shared<SubscriptionHandler<modulo_new_core::EncodedState>>(message_pair);
+        auto subscription = NodeT::template create_subscription<modulo_new_core::EncodedState>(
+            topic_name, this->qos_, subscription_handler->get_callback());
+        subscription_interface = subscription_handler->create_subscription_interface(subscription);
+        break;
+      }
+    }
+    this->inputs_.insert(std::make_pair(parsed_signal_name, subscription_interface));
```
eeberhard, line 520: I didn't think about this in the publisher, but we should check if the input / output already exists before doing all of this work. If the signal name already exists, `insert` will not change the map, but the subscription will have been created. We shouldn't allow creating two inputs with the same signal name, because ROS gets particular about creating and registering the same pubs / subs multiple times. That should be checked in the inputs_ map at the start of this function, before any of the subsequent logic. 

As a side note, I don't really like `insert(std::make_pair(..., ...))` and would prefer the more concise member initialiser list, available since C++11, as:
```suggestion
    this->inputs_.insert({parsed_signal_name, subscription_interface});
```
I know it's done the other way through the rest of the file so you don't have to change it, just pointing it out. And, where appropriate, the C++17 extensions `try_emplace` and `insert_or_assign` are preferable still. If you check for prior signal name existence in the map beforehand anyway, you could directly use `insert_or_assign(signal_name, interface)`.

eeberhard (approved):
Overall this looks really nice and is a big step indeed. I think we just need to be a bit careful about creating ROS interfaces that are not used (e.g. creating a subscription before trying to insert it into the map of inputs). 

If at some point we want re-assignable inputs / outputs, which would of course be an interesting feature, we just have to make sure to destroy the respective interface before calling `create_pub|sub`. So in any case, a check for the parsed signal name in the map of inputs or outputs before doing anything else is necessary, and we can worry about re-assignment at a later point.

Other than that it's hard to spot any other issues. Let's get it in there soon and start trying it out with some real components

domire8:
@eeberhard I was telling you about the tests that were failing and it seems we can't have a `ComponentInterfacePublicInterface` class in several test_*.cpp files. I would like to structure those test nicer in a subsequent PR, such that we can put templated interface classes (so we can test both node and lifecycle node) in a header and reuse them in different test files, do you think that would be possible? If yes, we should discuss how that could look like in practice

PR #65, opened 2022-04-29
Parameter translators in python

domire8:
Sadly there are no switch cases with our python version, so a lot of if conditions. I decided to list all possible cases to be precise and to not implicitely include possible new supported parameter types. All in all, the cpp behavior is quite well replicated, with `read_parameter` and `read_parameter_const`. The replicated tests pass exactly in the same way :+1:

### DIFF
--- a/source/modulo_new_core/test/python_tests/translators/test_parameters.py
+++ b/source/modulo_new_core/test/python_tests/translators/test_parameters.py
@@ -1,0 +1,130 @@
+import clproto
+import numpy as np
+import pytest
+import state_representation as sr
+from modulo_new_core.translators.parameter_translators import write_parameter, read_parameter, read_parameter_const
+from rclpy import Parameter
+
+
+def assert_np_array_equal(a: np.array, b: np.array, places=3):
+    try:
+        np.testing.assert_almost_equal(a, b, decimal=places)
+    except AssertionError as e:
+        pytest.fail(f'{e}')
+
+
+def test_parameter_write(parameters):
+    for name, value_types in parameters.items():
+        param = sr.Parameter(name, value_types[0], value_types[1])
+        ros_param = write_parameter(param)
+        assert ros_param.to_parameter_msg() == Parameter(name, value=value_types[0]).to_parameter_msg()
+
+
+def test_parameter_read_write(parameters):
+    for name, value_types in parameters.items():
+        ros_param = Parameter(name, value=value_types[0])
+        param = read_parameter(ros_param)
+        assert ros_param.name == param.get_name()
+        assert param.get_parameter_type() == value_types[1]
+        assert param.get_value() == value_types[0]
+
+        new_ros_param = write_parameter(param)
+        assert ros_param.to_parameter_msg() == new_ros_param.to_parameter_msg()
+
+
+def test_parameter_const_read(parameters):
+    for name, value_types in parameters.items():
+        param = sr.Parameter(name, value_types[0], value_types[1])
+        ros_param = Parameter(name, value=value_types[0])
+
+        new_param = read_parameter_const(ros_param, param)
+        assert new_param.get_name() == param.get_name()
+        assert new_param.get_parameter_type() == param.get_parameter_type()
+        assert new_param.get_value() == param.get_value()
+
+        new_param.set_name("other")
+        assert new_param.get_name() != param.get_name()
+
+
+def test_parameter_non_const_read(parameters):
+    for name, value_types in parameters.items():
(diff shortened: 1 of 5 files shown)


source/modulo_new_core/test/python_tests/translators/test_parameters.py
```diff
@@ -0,0 +1,130 @@
...
+
+        param_ref = param
+        new_param = read_parameter(ros_param, param)
+        assert new_param.get_name() == ros_param.name
+        assert new_param.get_parameter_type() == value_types[1]
+        assert new_param.get_value() == value_types[0]
+
+        assert param_ref.get_name() == ros_param.name
+        assert param_ref.get_parameter_type() == value_types[1]
+        assert param_ref.get_value() == value_types[0]
```
eeberhard, line 62: This is not quite equivalent to the C++ test. You should not construct the `param` with the actual value, otherwise the equality of `new_param == param_ref` is guaranteed, since `new_param == param`. 

I don't know if there is a default constructor in Python, but otherwise maybe you could try to pass `None` as the value, to ensure that `new_param != param` after the read.

```python
param = sr.Parameter(name, None, value_types[1])
```


For reference, in C++ it is:
```c++
// param value is default-constructed
auto param = std::make_shared<state_representation::Parameter<TypeParam>>("test");
// ros param has non-default value
rclcpp::Parameter ros_param("test", value);

// parameter reference copy has default value
auto param_ref = param;

// the parameter read updates the default value to the non-default
read_parameter(ros_param, param);
```

source/modulo_new_core/test/python_tests/translators/test_parameters.py
```diff
@@ -0,0 +1,130 @@
...
+        assert ros_param.name == param.get_name()
+        assert clproto.decode(clproto.from_json(ros_param.value)).get_name() == value_types[0].get_name()
+
+
+def test_state_parameter_read_write(state_parameters):
+    for name, value_types in state_parameters.items():
+        json = clproto.to_json(clproto.encode(value_types[0], value_types[2]))
+        ros_param = Parameter(name, value=json)
+        param = read_parameter(ros_param)
+        assert ros_param.name == param.get_name()
```
eeberhard, line 78: Super minor suggestion for readability:
```suggestion
        assert param.get_name() == ros_param.name
```

source/modulo_new_core/test/python_tests/translators/test_parameters.py
```diff
@@ -0,0 +1,130 @@
...
+        assert param_ref.get_parameter_type() == value_types[1]
+        assert param_ref.get_value() == value_types[0]
+
+
+def test_state_parameter_write(state_parameters):
+    for name, value_types in state_parameters.items():
+        param = sr.Parameter(name, value_types[0], sr.ParameterType.STATE, value_types[1])
+        ros_param = write_parameter(param)
+        assert ros_param.name == param.get_name()
+        assert clproto.decode(clproto.from_json(ros_param.value)).get_name() == value_types[0].get_name()
```
eeberhard, line 70: Could you add the `data()` equality check here too just to be comprehensive?

source/modulo_new_core/test/python_tests/translators/test_parameters.py
```diff
@@ -0,0 +1,130 @@
...
+
+    new_param = read_parameter(ros_param)
+    assert new_param.get_name() == ros_param.name
+    assert new_param.get_parameter_type() == sr.ParameterType.DOUBLE_ARRAY
+    assert new_param.get_value() == ros_mat.tolist()
+
+    new_param = read_parameter_const(ros_param, param)
+    assert new_param.get_name() == ros_param.name
+    assert new_param.get_parameter_type() == sr.ParameterType.MATRIX
+    assert_np_array_equal(new_param.get_value().flatten(), mat.flatten())
```
eeberhard, line 130: In C++ I did an equality check between the matrices, which also checks their shapes and not just their data (`EXPECT_EQ(new_mat, mat)`). Checking data equality after flattening does not properly test that the matrix was "unflattened" correctly. For this `np.array_equal(new_param.get_value(), mat)` might be preferable, or else (if floating point tolerance is an issue) just add a second assertion for `new_param.get_value().shape == mat.shape`

eeberhard (approved):
Looks great! I appreciate how you structured it.

I have a few comments for improving the tests, but if everything is still passing after those revisions then it's good to go imo.

domire8:
Thanks, I think I updated everything

PR #66, opened 2022-05-03
Improve test coverage in modulo new core

domire8:
Slightly over the limit of 250 lines due to another big switch case for the reading of a shared pointer of state.

The main change is the test coverage of the Publisher and Subscriber interface with 2 nodes communicating and checking that the messages have been received for all supported message types

### DIFF
--- a/source/modulo_new_core/test/cpp_tests/communication/test_communication.cpp
+++ b/source/modulo_new_core/test/cpp_tests/communication/test_communication.cpp
@@ -1,9 +1,5 @@
 #include <gtest/gtest.h>
 
-#include <chrono>
-#include <memory>
-#include <string>
-
 #include "rclcpp/rclcpp.hpp"
 
 #include "modulo_new_core/communication/MessagePair.hpp"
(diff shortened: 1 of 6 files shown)


source/modulo_new_core/test/cpp_tests/communication/test_communication.cpp
```diff
@@ -0,0 +1,141 @@
...
+      const std::shared_ptr<MessagePairInterface>& sub_message
+  ) {
+    pub_node_ = std::make_shared<MinimalPublisher<MsgT>>(topic_name, pub_message);
+    sub_node_ = std::make_shared<MinimalSubscriber<MsgT>>(topic_name, sub_message);
+    exec_->add_node(pub_node_);
+    exec_->add_node(sub_node_);
+  }
+
+  template<typename MsgT>
+  void add_nodes(
```
eeberhard, line 78: this signature is unused at the moment right? could you think of a simple test case that would use a callback instead of a shared value? It could also be added in a later PR

eeberhard (approved):
Sorry for the delayed review. This is a really nice addition, especially the modern usage of futures in the communication tester. Well designed!

source/modulo_new_core/test/cpp_tests/communication/test_communication.cpp
domire8: Good catch, no it's not used, it's rather a development artifact. Well I'm sure you could find a test case for that but we are already providing a custom callback to the subscription handler because it contains the future, so I think that's enough for now. Otherwise you'd need to create another future somewhere.

domire8: I removed it for now

PR #67, opened 2022-05-12
Add option to add deamon and update step function

domire8:
See commit messages for the changes in this PR. Review by commit

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -174,6 +174,11 @@ protected:
       bool fixed_topic = false, const std::string& default_topic = ""
   );
 
+  /**
+   * @brief Function to daemonize the callback function given in input.
+   * @param name The name of the daemon
+   * @param callback The callback of the daemon
+   */
   void add_daemon(const std::string& name, const std::function<void(void)>& callback);
 
   /**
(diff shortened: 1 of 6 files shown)


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -601,6 +599,44 @@ inline state_representation::CartesianPose ComponentInterface<NodeT>::lookup_tra
...
+                                   "No publisher for predicate " << predicate.first << " found.");
+      return;
+    }
+    predicate_publishers_.at(predicate.first)->publish(msg);
+  }
+}
+
+template<class NodeT>
+inline void ComponentInterface<NodeT>::publish_outputs() {
+  for (const auto& publisher: this->outputs_) {
```
eeberhard, line 618: For this function and the daemon call-backs, when you need to access both elements of the pair, I would prefer to use structured bindings

```c++
for (const auto& [signal, publisher]: this->outputs_)
```

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -241,6 +241,12 @@ class ComponentInterface : public NodeT {
 
   void step();
 
+  void publish_predicates();
```
eeberhard, line 244: docstrings for the new methods

eeberhard (commented):
Very nice overall to tie together the step function. Minor comments only.

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -290,56 +326,28 @@ ComponentInterface<NodeT>::ComponentInterface(
...
-    RCLCPP_DEBUG_STREAM(this->get_logger(), "Predicate " << name << " already exists, overwriting.");
-  } else {
-    this->predicate_publishers_.insert_or_assign(
-        name, this->template create_publisher<std_msgs::msg::Bool>(
-            this->generate_predicate_topic(name), 10
-        ));
-  }
-  this->predicates_.insert_or_assign(name, predicate);
+inline void ComponentInterface<NodeT>::step() {
+  this->publish_predicates();
```
domire8, line 330: As we just discussed, this is not correct. For a lifecycle component, publishing should only happen if the component is active

domire8: Should be good now

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -174,6 +179,13 @@ class ComponentInterface : public NodeT {
       bool fixed_topic = false, const std::string& default_topic = ""
   );
 
+  /**
+   * @brief Function to daemonize the callback function given in input.
```
eeberhard, line 183: Just a suggestion to improve this docstring
```suggestion
   * @brief Add a periodic callback function.
   * @details The provided function is evaluated periodically at the component step period. 
```
And, as discussed, perhaps `add_periodic_callback` is better than `add_daemon` as a function name.

eeberhard (approved):

PR #68, opened 2022-05-12
Implement execution thread for component

domire8:
This PR kind of concludes the implementation of the one shot component in CPP with the addition of the execution thread and the option to start the thread either on construction of on call of `start_thread`. Additionally, the component has the predicates `in_error_state` and `is_finished`

### DIFF
--- a/source/modulo_components/src/Component.cpp
+++ b/source/modulo_components/src/Component.cpp
@@ -22,7 +22,6 @@ void Component::start_thread() {
   }
   this->started_ = true;
   this->run_thread_ = std::thread([this]() { this->run(); });
-  this->run_thread_.join();
 }
 
 void Component::run() {


source/modulo_components/src/Component.cpp
```diff
@@ -4,7 +4,48 @@ using namespace modulo_new_core::communication;
...
+
+void Component::start_thread() {
+  if (this->started_) {
+    RCLCPP_ERROR_STREAM_THROTTLE(this->get_logger(), *this->get_clock(), 1000,
+                                 "Run thread for component '" << this->get_name() << "has already been started");
+    return;
+  }
+  this->started_ = true;
+  this->run_thread_ = std::thread([this]() { this->run(); });
+  this->run_thread_.join();
```
eeberhard, line 25: This will make `start_thread` a blocking function, at which point it is equivalent to calling `run()` directly. We shouldn't block the constructor from finishing if `start_thread` is true.
```suggestion
```

source/modulo_components/include/modulo_components/Component.hpp
```diff
@@ -16,9 +17,13 @@ class Component : public ComponentInterface<rclcpp::Node> {
    * @brief Constructor from node options
    * @param node_options node options as used in ROS2 Node
    */
-  explicit Component(const rclcpp::NodeOptions& node_options);
+  explicit Component(const rclcpp::NodeOptions& node_options, bool start_thread = true);
 
 protected:
+  void start_thread();
```
eeberhard, line 23: docstrings on all the new functions please

eeberhard (commented):
Nice addition. Just a few comments!

eeberhard (approved):
Thanks for the changes

PR #69, opened 2022-05-13
Clean up python message translators

domire8:
Several small touchups in the python message translators as well as renaming to the same name as in CPP. Addition of docstrings.


source/modulo_new_core/modulo_new_core/translators/message_readers.py
```diff
@@ -0,0 +1,102 @@
...
+import sensor_msgs.msg
+import sensor_msgs.msg
+import state_representation as sr
+from modulo_new_core.encoded_state import EncodedState
+
+MsgT = TypeVar('MsgT')
+StateT = TypeVar('StateT')
+
+
+def __read_xyz(msg: Union[geometry.Point, geometry.Vector3]) -> List[float]:
```
eeberhard, line 15: I know you didn't write this in this PR, but I'm not sure what benefit if any the `__` prefix brings here (and in the other functions in this file and the next). If the intent is to hide them from being importable, I don't think any name mangling will happen since this is not a class methods. But, in general, this is a useful function and can/should have "public" visibility if anyone did care to import it.

eeberhard (commented):
This looks very tidy! I can't check the tests at the moment but it's a nice improvement.

source/modulo_new_core/modulo_new_core/translators/message_readers.py
domire8: I guess I consider them as helper functions and would probably make them private in cpp but yeah I agree that this is nice to have public. We won't be able to duplicate cpp exactly anyway.

eeberhard (approved):

PR #70, opened 2022-05-16
Add python component with execution thread

domire8:
As title suggests

### DIFF
--- a/source/modulo_components/modulo_components/component.py
+++ b/source/modulo_components/modulo_components/component.py
@@ -1,0 +1,72 @@
+from threading import Thread
+
+from modulo_components.component_interface import ComponentInterface
+
+
+class Component(ComponentInterface):
+    """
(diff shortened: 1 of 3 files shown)


source/modulo_components/modulo_components/component.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Constructs all the necessary attributes and declare all the parameters.
+            Parameters:
+                node_name (str): name of the node to be passed to the base Node class
+                start_thread (bool): start the execution thread at construction
+        """
+        super().__init__(node_name, *kargs, **kwargs)
+        self.__started = False
+        self.__run_thread = None
+        self.add_predicate("in_error_state", False)
```
eeberhard, line 26: As we discussed just now, this predicate can be added in the base ComponentInterface

eeberhard (approved):
Seems good

PR #71, opened 2022-05-17
Improve components tests

domire8:
With this PR, I'm forward declaring our friend test classes such that we can use `TYPED_TEST` with `NodeT` template test classes to test both lifecycle and normal component interface.

A consequence of this is that I needed to add `this->` in the parameter tests, which results in a big diff.

Also added a few more test cases


eeberhard:
The forward declared templates are very nice. However, I don't particularly the like the multiple friend classes in the component interface. I would prefer to keep that as a single class and just expose as much as we need. Otherwise if we want to test more stuff we might end up having to forward declare many friend test classes.

I also think that we might want to re-use the interface classes across test source files, so they could be declared in a separate header file.

I made these changes on a suggestion branch, let me know what you think:
feature/improve-components-tests-suggestion

domire8:
Very nice, thanks! Good to go?

eeberhard (approved):

PR #72, opened 2022-05-20
Minor improvements

domire8:
A few small things that came up during our discussion yesterday that I wanted to fix in the C++ components before I dig deeper into the bigger topics (namely exception handling and lifecycle transitions).

One important thing to take away from this PR is that we hide all of NodeT methods except a small set that is actually required (name, clock, logger, node_base_interface)

Review by commit, they are small and easy.


eeberhard (approved):
👍 very good incremental improvements, thanks!

PR #73, opened 2022-05-24
Parameters in python ComponentInterface

domire8:
Instead of using a parameter map like in C++, I use here a dict to store either the attribute name or the actual parameter object as the parameters of the component interface. By storing the attribute name and then get/set the parameter via `__getattribute__` and `__setattr__` we have a similar behavior as in C++ with the pointers.

Another important point to note is that I cannot override `get_parameter` as it is used by rclpy to get parameters too. Hence, I check from which file `get_parameter` is called with `co_filename = sys._getframe().f_back.f_code.co_filename` and if it contains `rclpy`, the Node.get_parameter is called. This is probably not best practice but it may work...

Last but not least, the `read_only` behavior from C++ couldn't be replicated yet, I'm not sure if that feature exists in rclpy too. I marked those parts as TODO.


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -67,29 +67,159 @@ def __step(self):
...
+
+        :param parameter: Either the name of the parameter attribute or the parameter itself
+        :param description: The parameter description
+        :param read_only: If True, the value of the parameter cannot be changed after declaration
+        """
+        try:
+            if isinstance(parameter, sr.Parameter):
+                sr_parameter = parameter
+            elif isinstance(parameter, str):
+                sr_parameter = self.__getattribute__(parameter)
```
eeberhard, line 94: It seems we should also check for the type of the attribute, otherwise `write_parameter` or `sr_parameter.get_name()` could fail. 

```suggestion
                attr = self.__getattribute__(parameter)
                if isinstance(attr, sr.Parameter):
                    sr_parameter = attr
                else:
                    raise ...
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -67,29 +67,159 @@ def __step(self):
...
+    def set_parameter_value(self, name: str, value: T, parameter_type: sr.ParameterType) -> None:
+        """
+        Set the value of a parameter. The parameter must have been previously declared. If the parameter is an
+        attribute, the attribute is updated.
+
+        :param name: The name of the parameter
+        :param value: The value of the parameter
+        :param parameter_type: The type of the parameter
+        """
+        result = self.set_parameters([write_parameter(sr.Parameter(name, value, parameter_type))])[0]
```
eeberhard, line 160: I think this would be more readable in multiple lines.

In addition, we could add an explicit check for if parameter exists before we go through the effort of trying to set it.

```python
if name not in self._parameter_dict.keys():
    raise ComponentParameterError(f"Failed to get parameter '{name}'")

ros_param = write_parameter(sr.Parameter(name, value, parameter_type))
result = self.set_parameters([ros_param])[0]
```

eeberhard (commented):
Sorry for the late review. This looks very good as a counterpart to the C++ component parameters, with some language-specific solutions. The dictionary of parameters with get/set attribute is great. The rclpy path is a decent workaround for the time being, but probably not robust long-term. Just a few comments.

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -67,29 +67,165 @@ def __step(self):
...
+        """
+        return f"/predicates/{self.get_name()}/{predicate_name}"
+
+    def add_parameter(self, parameter: Union[str, sr.Parameter], description: str, read_only=False) -> None:
+        """
+        Add a parameter. This method stores either the name of the attribute corresponding to the parameter object or
+        a parameter object directly in the local parameter dictionary and declares the equivalent ROS parameter on the
+        ROS interface. If an attribute name is provided, changing the ROS parameter will also update the provided
+        attribute and vice versa. If the provided argument is not an attribute name or a Parameter object, nothing
+        happens.
```
eeberhard, line 84: Ah, in my first comment I was just adjusting the grammar of this sentence, but actually it's not quite correct. If the argument is not valid, it throws.

domire8: Ah yeah right but it will actually not throw anymore once i resolve the TODO below. I will revise all that in a next PR

eeberhard (approved):

PR #77, opened 2022-06-12
TF improvements

domire8:
The `add_tf_xxx` and send/lookup transform was already here but not tested and exceptions not caught according to the guidelines. I'm fixing this here and doing the same in python, where it was not implemented so far.

I changed from tf broadcaster to a publisher for the `tf` topic, because I was not able to create a tf broadcaster (probably due to private inheritance)

Review by commit

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -707,7 +707,7 @@ inline void ComponentInterface<NodeT>::add_tf_listener() {
 
 template<class NodeT>
 inline void ComponentInterface<NodeT>::activate_tf_broadcaster() {
-  if (this->publisher_type_ == modulo_new_core::communication::PublisherType::PUBLISHER) {
+  if (this->publisher_type_ != modulo_new_core::communication::PublisherType::LIFECYCLE_PUBLISHER) {
     return;
   }
   try {
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -336,6 +336,7 @@ private:
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_; ///< TF buffer
   std::shared_ptr<tf2_ros::TransformListener> tf_listener_; ///< TF listener
   std::shared_ptr<rclcpp::Publisher<tf2_msgs::msg::TFMessage>> tf_broadcaster_; ///< TF broadcaster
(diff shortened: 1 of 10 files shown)


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -674,7 +685,8 @@ template<class NodeT>
 inline void ComponentInterface<NodeT>::add_tf_broadcaster() {
   if (this->tf_broadcaster_ == nullptr) {
     RCLCPP_DEBUG(this->get_logger(), "Adding TF broadcaster.");
-    this->tf_broadcaster_ = std::make_shared<tf2_ros::TransformBroadcaster>(this->shared_from_this());
+    console_bridge::setLogLevel(console_bridge::CONSOLE_BRIDGE_LOG_NONE);
```
eeberhard, line 689: I take it this is used to suppress TF errors / warnings outside of the ros logger context?

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -684,46 +696,81 @@ template<class NodeT>
...
     this->tf_buffer_ = std::make_shared<tf2_ros::Buffer>(this->get_clock());
     this->tf_listener_ = std::make_shared<tf2_ros::TransformListener>(*this->tf_buffer_);
   } else {
     RCLCPP_DEBUG(this->get_logger(), "TF buffer and listener already exist.");
   }
 }
 
+template<class NodeT>
+inline void ComponentInterface<NodeT>::activate_tf_broadcaster() {
+  if (this->publisher_type_ == modulo_new_core::communication::PublisherType::PUBLISHER) {
```
eeberhard, line 710: Small nitpick but I think this may be safer / more readable
```suggestion
  if (this->publisher_type_ != modulo_new_core::communication::PublisherType::LIFECYCLE_PUBLISHER) {
```

eeberhard (approved):
Looks good to me overall

source/modulo_components/include/modulo_components/ComponentInterface.hpp
domire8: yes indeed, if we would fork rviz and do the same there, we wouldn't get any of these annoying logs anymore

domire8:
> It seems all good for me. One thing we might consider is to add a `add_static_tf_broadcaster` that works in exactly the same way but publishes on `tf_static` and not `tf`. If you add that you might want to directly add it with a state on construction and remove possibilities to publish periodically from it.

Indeed that could be something we can add as well, I added a TODO for that

PR #78, opened 2022-06-13
Add inputs and outputs to python component interface

domire8:
This PR finally adds inputs and outputs to the component interface in python. It also includes a few minor changes in the translators to make this feature possible. I also changed `add_periodic_function` to `add_periodic_callback` and updated the logging in a few places.


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -278,6 +284,86 @@ def set_predicate(self, predicate_name: str, predicate_value: Union[bool, Callab
...
+                raise AddSignalError(f"Output with name '{parsed_signal_name}' already exists")
+            topic_name = default_topic if default_topic else "~/" + parsed_signal_name
+            self.add_parameter(sr.Parameter(parsed_signal_name + "_topic", topic_name, sr.ParameterType.STRING),
+                               f"Output topic name of signal '{parsed_signal_name}'", fixed_topic)
+            translator = None
+            if message_type == Bool or message_type == Float64 or \
+                    message_type == Float64MultiArray or message_type == Int32 or message_type == String:
+                translator = modulo_writers.write_std_msg
+            elif message_type == EncodedState:
+                # TODO switch case to find message type
```
domire8, line 302: This TODO is actually quite important and something we need to discuss. I can see these two options, there are probably others:
- actually implement the switch case
- just add another optional argument that would be used here

The drawback of the second option is that we put the responsibility on the user to provide the correct clproto message type. The drawback of the first option is that it will result in a lot of nested `isinstance` conditions, where I'm not sure if its very robust.

domire8:
Closing this in favor of more compact PRs, this one got a bit out of control

PR #81, opened 2022-06-17
Add QoS and periodic callbacks to Python ComponentInterface

domire8:
A short neat PR with a few simple updates to the ComponentInterface. Review by commit

### DIFF
--- a/source/modulo_components/modulo_components/component_interface.py
+++ b/source/modulo_components/modulo_components/component_interface.py
@@ -14,6 +14,7 @@ from rcl_interfaces.msg import SetParametersResult
 from rclpy.duration import Duration
 from rclpy.node import Node
 from rclpy.publisher import Publisher
+from rclpy.qos import QoSProfile
 from rclpy.time import Time
 from std_msgs.msg import Bool
 from tf2_ros import TransformBroadcaster
@@ -33,6 +34,7 @@ class ComponentInterface(Node):
         predicate_publishers (dict(rclpy.Publisher(Bool))): map of publishers associated to each predicate.
         parameter_dict: dict of parameters
         periodic_callbacks: dict of periodic callback functions
+        qos: the Quality of Service for publishers and subscribers
         tf_buffer (tf2_ros.Buffer): the buffer to lookup transforms published on tf2.
         tf_listener (tf2_ros.TransformListener): the listener to lookup transforms published on tf2.
         tf_broadcaster (tf2_ros.TransformBroadcaster): the broadcaster to publish transforms on tf2
@@ -54,6 +56,8 @@ class ComponentInterface(Node):
         self.__tf_buffer: Optional[Buffer] = None
         self.__tf_listener: Optional[TransformListener] = None
         self.__tf_broadcaster: Optional[TransformBroadcaster] = None
+
+        self._qos = QoSProfile(depth=10)
 
         self.add_on_set_parameters_callback(self.__on_set_parameters_callback)
         self.add_parameter(sr.Parameter("period", 0.1, sr.ParameterType.DOUBLE),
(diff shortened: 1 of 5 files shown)


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -32,6 +32,7 @@ class ComponentInterface(Node):
         predicates (dict(Parameters(bool))): map of predicates added to the Component.
         predicate_publishers (dict(rclpy.Publisher(Bool))): map of publishers associated to each predicate.
         parameter_dict: dict of parameters
+        periodic_callbacks: dict of periodic callback functions
```
buschbapti, line 37: Can you add `(dict(Callable))` and `dict(Parameters)` the line above just to make it consistent? (Sorry I can't make suggestions it seems.

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -337,3 +339,30 @@ def lookup_transform(self, frame_name: str, reference_frame_name="world", time_p
...
+        Add a periodic callback function. The provided function is evaluated periodically at the component step period.
+
+        :param name: The name of the callback
+        :param callback: The callback function that is evaluated periodically
+        """
+        if not name:
+            self.get_logger().error("Failed to add periodic function: Provide a non empty string as a name.")
+            return
+        if name in self._periodic_callbacks.keys():
+            self.get_logger().debug(f"Periodic function '{name}' already exists, overwriting.")
```
buschbapti, line 359: I would put this at info level or even maybe warn (I let you decide). It seems too important to notify to keep a debug log.

source/modulo_components/src/LifecycleComponent.cpp
```diff
@@ -15,11 +15,17 @@ LifecycleComponent::LifecycleComponent(const rclcpp::NodeOptions& node_options)
...
-    this->on_step();
+    if (this->get_predicate("is_active")) {
+      this->publish_outputs();
+      this->evaluate_periodic_callbacks();
+      this->on_step();
+    }
+  } catch (const std::exception& ex) {
+    RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to execute step function:" << ex.what());
+    // TODO handle error in lifecycle component
+//    this->raise_error();
```
buschbapti, line 28: why doesn't that raise the error anymore?

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -49,37 +53,26 @@ def __init__(self, node_name: str, *kargs, **kwargs):
         self._parameter_dict: Dict[str, Union[str, sr.Parameter]] = {}
         self._predicates: Dict[str, Union[bool, Callable[[], bool]]] = {}
         self._predicate_publishers: Dict[str, Publisher] = {}
+        self._periodic_callbacks: Dict[str, Callable[[], None]] = {}
         self.__tf_buffer: Optional[Buffer] = None
         self.__tf_listener: Optional[TransformListener] = None
         self.__tf_broadcaster: Optional[TransformBroadcaster] = None
 
+        self._qos = QoSProfile(depth=10)
```
buschbapti, line 61: Is this overwriting a base class variable? Because it seems to never been used anywhere so just wondering how it affect the rest.

buschbapti (commented):
Seems quite good and handy. A few minor comments.

source/modulo_components/modulo_components/component_interface.py
domire8: It's going to be used once I add publishers and subscribers in one of my next PRs

source/modulo_components/modulo_components/component_interface.py
domire8: done :+1:

source/modulo_components/modulo_components/component_interface.py
domire8: I actually don't fully agree, and I'd rather remove the type hints in the docstring because I have them now very nice when I define the attribute (check a few lines further down). Also `dict(Parameters)` is not technically correct because it can be a parameter or it can be the name of an attribute that refers to the parameter...
I think suggestions don't work if you review by commit

domire8: I added it for now, we have to figure out how we can do this properly later

buschbapti (approved):

PR #82, opened 2022-06-17
Fix MessagePair to enable creating from any SR type

domire8:
There was a rather big issue in the current MessagePair implementation: Due to the fact that all 6 `make_shared_message_pair<MsgT, Data>` were specialized in the source file, it was not possible to create message pairs with
```
make_shared_message_pair<state_representation::State, EncodedState>(std::shared_ptr<CartesianTwist> data)
```
(i.e. you could only make a pair if the input data pointer was also a `State` and not one of its derived classes). This is fixed now by keeping one `make_shared_message_pair` templated (props to @buschbapti the templating guru).

Note that without the `inline` keywords in the header, this solution wouldn't work (just for future reference, the `inline` keyword can actually be really important)

### DIFF
--- a/source/modulo_new_core/include/modulo_new_core/communication/MessagePair.hpp
+++ b/source/modulo_new_core/include/modulo_new_core/communication/MessagePair.hpp
@@ -106,7 +106,41 @@ inline void MessagePair<MsgT, DataT>::set_data(const std::shared_ptr<DataT>& dat
 }
 
 template<typename DataT>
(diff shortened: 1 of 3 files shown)


buschbapti (approved):
Nice! And I also like my new title

eeberhard:
> Nice! And I also like my new title

I guess I should concede my role of CTO to @buschbapti - Chief Template Officer

PR #83, opened 2022-06-20
Add inputs and outputs to python component interface

domire8:
Let's try this again. This PR adds the input and output signals to the python component interface. As with the parameters, it takes a name of a class attribute that should be used to publish messages and receive messages.

### DIFF
--- a/source/modulo_components/modulo_components/component_interface.py
+++ b/source/modulo_components/modulo_components/component_interface.py
@@ -28,7 +28,6 @@ from tf2_ros import TransformBroadcaster
 from tf2_ros.buffer import Buffer
 from tf2_ros.transform_listener import TransformListener
 
-DataT = TypeVar('DataT')
 MsgT = TypeVar('MsgT')
 T = TypeVar('T')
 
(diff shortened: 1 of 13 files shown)


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -273,6 +285,106 @@ def set_predicate(self, name: str, value: Union[bool, Callable[[], bool]]):
...
+        """
+        try:
+            parsed_signal_name = parse_signal_name(signal_name)
+            if not parsed_signal_name:
+                raise AddSignalError(f"Failed to add input '{signal_name}': Parsed signal name is empty.")
+            if parsed_signal_name in self._inputs.keys():
+                raise AddSignalError(f"Failed to add input '{parsed_signal_name}': Input already exists")
+            topic_name = default_topic if default_topic else "~/" + parsed_signal_name
+            self.add_parameter(sr.Parameter(parsed_signal_name + "_topic", topic_name, sr.ParameterType.STRING),
+                               f"Output topic name of signal '{parsed_signal_name}'", fixed_topic)
```
eeberhard, line 362: Should be "Input topic name..."

eeberhard (approved):
Sorry for the delayed review. I wanted to make sure I gave it enough time and attention.

The output publisher is a really nice showcase of the abstraction you built, with the message and translator attributes. Great work!

PR #84, opened 2022-06-20
Inherit from NodeT with public keyword

domire8:
One of my initial tests showed that it is not possible to inherit with `private` from `NodeT` and we were already thinking of switching back to `public` so here it is. As a consequence, we can also switch back to a `TransformBroadcaster` and create `rclcpp::Publisher` explicitly for all predicate publishers. And finally, we can use lambda functions with `get_current_state` in the lifecycle components to get the current state for all predicates.

An important TODO here will be to hide all non necessary methods from ROS in the components

### DIFF
--- a/source/modulo_components/src/LifecycleComponent.cpp
+++ b/source/modulo_components/src/LifecycleComponent.cpp
@@ -8,10 +8,26 @@ namespace modulo_components {
 
 LifecycleComponent::LifecycleComponent(const rclcpp::NodeOptions& node_options) :
     ComponentInterface<rclcpp_lifecycle::LifecycleNode>(node_options, PublisherType::LIFECYCLE_PUBLISHER) {
-  this->add_predicate("is_unconfigured", true);
-  this->add_predicate("is_inactive", false);
-  this->add_predicate("is_active", false);
-  this->add_predicate("is_finalized", false);
+  this->add_predicate(
+      "is_unconfigured", [this] {
+        return this->get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_UNCONFIGURED;
+      }
+  );
+  this->add_predicate(
+      "is_inactive", [this] {
+        return this->get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_INACTIVE;
+      }
+  );
+  this->add_predicate(
+      "is_active", [this] {
+        return this->get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_ACTIVE;
+      }
+  );
+  this->add_predicate(
+      "is_finalized", [this] {
+        return this->get_current_state().id() == lifecycle_msgs::msg::State::PRIMARY_STATE_FINALIZED;
+      }
+  );
 }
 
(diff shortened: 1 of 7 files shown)


source/modulo_components/src/LifecycleComponent.cpp
```diff
@@ -203,11 +203,9 @@ LifecycleComponent::on_error(const rclcpp_lifecycle::State& previous_state) {
   if (!error_handled) {
     RCLCPP_ERROR(get_logger(), "Error processing failed! Entering into the finalized state.");
     // TODO: reset and finalize all needed properties
-    this->set_predicate("is_finalized", true);
     return rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn::ERROR;
   }
   this->set_predicate("in_error_state", false);
```
buschbapti, line 208: The `in_error_state` could also be using the `onError` lifecycle state (cf. https://design.ros2.org/articles/node_lifecycle.html). We need to find how to properly trigger that transition, Doing that also allow to overwrite the `onError()` function that can perform some recovery mechanisms to put back the node in unconfigured if successful.

buschbapti (approved):
Nice stuff especially with the way predicates are handled. See comment for future TODO.

source/modulo_components/src/LifecycleComponent.cpp
eeberhard: As you can see from this function, we do already have the on_error callback here that can attempt recovery to unconfigured. It is triggered when the other lifecycle transition callbacks return ERROR. It is triggered by the lifecycle state machine internally. I also want raise_error() to invoke this callback, though that is not yet implemented. 

I don't think that a lambda function on get_state should be used for the error state, because it's sort of a special case that is disassociated from the lifecycle states. The lifecycle error state is always transient, and goes into unconfigured or finalised on its own. If it goes to unconfigured, then it's fair to say it's not in error state. And, if it goes to finalised without ever having an error, that's also fine. But if it goes to finalised because of an error, then the application (state engine) must know that distinction. The difference is basically [finalised AND NOT error] vs [finalised AND error]. Therefore I would rather set the "in_error_state" predicate to `true` instead of a state-based dynamic value, so that if we go to finalised, we continue to be in "error state" and can do appropriate transitions from the application perspective (e.g. unload the component or trigger something else)

eeberhard (approved):
Nice work. I know we had some back and forth on this topic and the private inheritance seemed like a smart solution at first. Thanks for doing the work to bring it back and implementing the suggestions for predicates!

source/modulo_components/src/LifecycleComponent.cpp
buschbapti: Indeed that's clear thanks for the explanation. The `raise_error()`  that invoke the transition would still be nice and should be marked as a definite todo. I am fine with your explanation concerning the predicate then.

domire8: The TODO has already been at the bottom of LifecycleComponent.hpp for few PRs, I will try to tackle that soon.

PR #85, opened 2022-06-22
Make sure predicate is a bool

domire8:
After testing with some base components, I had to learn the hard way that after some operations with state representation types, e.g. 
```
collision = target_tf.get_position()[2] - center_tf.get_position()[2] < 0
self._in_bounds = not collision if self._flip_normal.get_value() else collision
```
the resulting variable could be of type `numpy.bool` or `numpy.bool_` and I think we should catch those cases in the component interface and convert to builtin bool.


domire8:
After discussions, it is probably better not to check for certain types but to catch the AssertionError and log a more informative error message.

PR #86, opened 2022-06-27
Translation of empty parameters

domire8:
Now that the default value of empty parameters was fixed in control libraries PR [#298](https://github.com/epfl-lasa/control-libraries/pull/298), I can use the `galactic-devel` image of ros2-control-libraries and make sure that the translation between state representation parameters and ROS parameters works correctly. What happened before was:
- an empty state rep Parameter of type double, int, etc would have a non-default value (for example 1 for double).
- an empty state rep Parameter of type XXX_ARRAY would have ROS parameter type BYTE_ARRAY

Both of these bugs are fixed now and you get the correct parameter type in ROS with correct default value.


buschbapti (approved):
Nice catch and fix

eeberhard (approved):
Looks good. This just means we will need to release CL 6.1 before the modulo refactor

PR #87, opened 2022-06-28
Add fallback name to constructors

domire8:
A small addition that I like because it allows to give a component a default name if it's not provided through the node options


buschbapti (approved):

eeberhard (approved):
Nice, makes sense. I think `fallback` is the correct term here since `component_name` or `default_name` would imply it's always set unless redefined.

PR #91, opened 2022-06-30
Remove old modulo_core and rename modulo_new_core

domire8:
As the title states


source/modulo_components/package.xml
```diff
@@ -4,22 +4,17 @@
   <name>modulo_components</name>
   <version>0.0.1</version>
   <description>TODO</description>
-  <maintainer email="baptiste@aica.tech">TODO</maintainer>
```
eeberhard, line 7: Haha these are the laziest TODOs I've ever seen, thanks for spotting and resolving them!

eeberhard (approved):
Looks clean! As is often the case, good refactors do not shy away from deleting 100 lines of code for 10 new ones. A respectful send-off to Cell and other modulo inventions that laid the groundwork for our modular composition framework.

PR #92, opened 2022-06-30
Modulo Refactor

domire8:
Title doesn't matter much, because I assume we will do a fast forward such that all commits of the refactor branch are preserved right? 

All commits have been reviewed.


eeberhard:
🚀

PR #93, opened 2022-06-30
Allow declaring unintialized ROS parameters

domire8:
Ok wow that was painful. So far, it was not possible to create ROS parameters from an empty state rep parameter. One idea was to do that: 
```
if sr_parameter.is_empty():
    self.declare_parameter(ros_param.name, None,
                           descriptor=ParameterDescriptor(description=description,
                                                          type=Parameter.Type.DOUBLE_ARRAY))
```
This however, does not actually create an uninitialized ROS parameter of type double array, but a ROS parameter of type `NOT_SET`. Which means that next time we want to set that parameter in the on_set_parameter_callback, it will be incompatible. So the solution is that:
```
if sr_parameter.is_empty():
    self.declare_parameter(ros_param.name, None,
                           descriptor=ParameterDescriptor(description=description,
                                                          type=Parameter.Type.NOT_SET.value,
                                                          dynamic_typing=True))
```
We declare the parameter with type NOT_SET and allow dynamic_typing. Now you might say, no wait we dont want dynamic typing. Well it's really only dynamically typed until its set for the first time because every following time that we want to modify the parameter, we are checking that the types correspond, so we are good. This behaviour should be illustrated in the test. We can discuss if thats not what we want.

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -400,9 +400,9 @@ inline void ComponentInterface<NodeT>::add_parameter(
       descriptor.description = description;
       descriptor.read_only = read_only;
       if (parameter->is_empty()) {
-        descriptor.dynamic_typing = true;
         NodeT::declare_parameter(
-            parameter->get_name(), utilities::get_ros_parameter_type(parameter->get_parameter_type()));
+            parameter->get_name(), utilities::get_ros_parameter_type(parameter->get_parameter_type()), descriptor
+        );
       } else {
         NodeT::declare_parameter(parameter->get_name(), ros_param.get_parameter_value(), descriptor);
       }
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -399,11 +399,22 @@ inline void ComponentInterface<NodeT>::add_parameter(
       rcl_interfaces::msg::ParameterDescriptor descriptor;
       descriptor.description = description;
       descriptor.read_only = read_only;
-      NodeT::declare_parameter(parameter->get_name(), ros_param.get_parameter_value(), descriptor);
+      if (parameter->is_empty()) {
+        descriptor.dynamic_typing = true;
+        NodeT::declare_parameter(
+            parameter->get_name(), utilities::get_ros_parameter_type(parameter->get_parameter_type()));
+      } else {
+        NodeT::declare_parameter(parameter->get_name(), ros_param.get_parameter_value(), descriptor);
+      }
     } else {
-      RCLCPP_WARN_STREAM(this->get_logger(),
-                         "Parameter '" << parameter->get_name() << "' already exists, overwriting.");
-      NodeT::set_parameter(ros_param);
+      if (parameter->is_empty()) {
+        RCLCPP_ERROR_STREAM(this->get_logger(),
+                            "Cannot overwrite parameter '" << parameter->get_name() << "' with an empty parameter.");
+      } else {
+        RCLCPP_WARN_STREAM(this->get_logger(),
+                           "Parameter '" << parameter->get_name() << "' already exists, overwriting.");
+        NodeT::set_parameter(ros_param);
+      }
     }
   } catch (const std::exception& ex) {
     RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to add parameter '" << parameter->get_name() << "': " << ex.what());
@@ -460,7 +471,7 @@ ComponentInterface<NodeT>::on_set_parameters_callback(const std::vector<rclcpp::
       auto new_parameter = modulo_core::translators::read_parameter_const(ros_parameter, parameter);
       if (!validate_parameter(new_parameter)) {
         result.successful = false;
-        result.reason += "Parameter " + ros_parameter.get_name() + " could not be set! ";
+        result.reason += "Validation of parameter '" + ros_parameter.get_name() + "' returned false!";
       } else {
         // update the value of the parameter in the map
         modulo_core::translators::copy_parameter_value(new_parameter, parameter);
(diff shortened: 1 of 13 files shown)


source/modulo_components/test/python/test_component.py
```diff
@@ -19,3 +22,19 @@ def test_add_output(component):
...
+def test_empty_parameter(component):
+    component.add_parameter(sr.Parameter("test", sr.ParameterType.DOUBLE_ARRAY), "description")
+    assert component.get_parameter("test").get_parameter_type() == sr.ParameterType.DOUBLE_ARRAY
+    assert Node.get_parameter(component, "test").type_ == Parameter.Type.NOT_SET
+
+    component.set_parameter_value("test", [1.0, 2.0], sr.ParameterType.DOUBLE_ARRAY)
+    assert Node.get_parameter(component, "test").type_ == Parameter.Type.DOUBLE_ARRAY
+
+    component.set_parameter_value("test", "test", sr.ParameterType.STRING)
+    assert component.get_parameter("test").get_parameter_type() == sr.ParameterType.DOUBLE_ARRAY
```
eeberhard, line 36: maybe you can add some simple comments here to make it clear what the test is expecting. In this case the parameter validation fails and the type (and value) are not changed, but it's not particularly obvious

eeberhard (commented):
That's a smart solution! Thanks for taking the time to work it out. 

The consequence is that "default empty" parameters cannot be validated automatically the first time they are set, which is acceptable but will need to be made clear in the documentation later. Any validation would have to occur "in place" where you get and use the parameter value. 

One question before I approve - what happens if we try to get the value of a parameter that is NOT_SET? Internally I suppose it's just the sr.Parameter object or attribute, which will be "empty". But if I were to read that parameter on the ROS interface, what would that look like?

domire8:
> One question before I approve - what happens if we try to get the value of a parameter that is NOT_SET? Internally I suppose it's just the sr.Parameter object or attribute, which will be "empty". But if I were to read that parameter on the ROS interface, what would that look like?

Internally, you get the sr.Parameter which is empty yes but you can still get the value, will be the default empty value of that specific parameter type.
If you do `ros2 param get /component pose_values` you get a message back that says `Parameter not set.`.

domire8:
> The consequence is that "default empty" parameters cannot be validated automatically the first time they are set, which is acceptable but will need to be made clear in the documentation later. Any validation would have to occur "in place" where you get and use the parameter value.

Quite true, maybe we can resolve that by moving around that `continue

buschbapti:
I am not sure exactly what was the initial issue (maybe I missed it if we discussed about it). But it seems what you wanted is to declare a parameter without giving it a default value, isn't it? Did you try the solution specified here https://github.com/ros2/rclcpp/blob/master/rclcpp/doc/notes_on_statically_typed_parameters.md ? I guess you might have but just in case to be sure.

domire8:
> I am not sure exactly what was the initial issue (maybe I missed it if we discussed about it). But it seems what you wanted is to declare a parameter without giving it a default value, isn't it? Did you try the solution specified here https://github.com/ros2/rclcpp/blob/master/rclcpp/doc/notes_on_statically_typed_parameters.md ? I guess you might have but just in case to be sure. Maybe something like:
> 
> ```
> self.declare_parameter(ros_param.name, rclpy::ParameterType::PARAMETER_DOUBLE_ARRAY, 
>                                       descriptor=ParameterDescriptor(description=description,
>                                                                                            type=Parameter.Type.DOUBLE_ARRAY))
> ```
> 
> I don't know if that would work in python though as I am not sure about the signature of the function there.

Yeahhh this is getting a bit annyoing, I will have to call a developer standup to clarify those issues...Python and C++ really don't behave the same here..

domire8:
So the current progress is as followed:
- In C++, adding an empty parameter from the modulo interface is okay, getting it from the ROS interface throws an exception, and setting a parameter with a  `NOT_SET` parameter on the ROS interface does undeclare that parameter
- In Python, adding an empty parameter from the modulo interface is okay, getting it from the ROS interface does not throw an exception and setting a parameter with a `NOT_SET` parameter on the ROS interface undeclares that parameter

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -399,11 +399,22 @@ inline void ComponentInterface<NodeT>::add_parameter(
       rcl_interfaces::msg::ParameterDescriptor descriptor;
       descriptor.description = description;
       descriptor.read_only = read_only;
-      NodeT::declare_parameter(parameter->get_name(), ros_param.get_parameter_value(), descriptor);
+      if (parameter->is_empty()) {
+        descriptor.dynamic_typing = true;
+        NodeT::declare_parameter(
+            parameter->get_name(), utilities::get_ros_parameter_type(parameter->get_parameter_type()));
```
buschbapti, line 405: Shouldn't `descriptor` be third argument there?

buschbapti, line 405: I was looking by commit but it seems it changed line. My comment was for here.

domire8: Oh that might solve my problem!

domire8: It doesn't...you cannot specify the type as second argument and allow dynamic typing at the same time

buschbapti (dismissed):
This looks good now and I believe well tested. Some minor style comments.

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -460,7 +471,7 @@ ComponentInterface<NodeT>::on_set_parameters_callback(const std::vector<rclcpp::
       auto new_parameter = modulo_core::translators::read_parameter_const(ros_parameter, parameter);
```
eeberhard, line 471: This is problematic when `parameter` is empty. Because the ros_parameter is of type string when you add_parameter, new_parameter will _not_ be empty. This means that `validate_parameter` will see a different, non-empty, default-valued version of the parameter. For an integer, this might mean that we want to add a "not set" number, but validate parameter will see a value of `0` instead.

eeberhard (commented):
As discussed, I think we have to revert to using the NOT_SET parameter type to make sure the parameter validation can be fully consistent.

domire8:
Close in favor of #97

PR #94, opened 2022-07-01
Improve exception handling in modulo_core

domire8:
In order to improve the exception handling in modulo core and components, I'm adding here in a first step a `CoreException` from which all exceptions in modulo core inherit. This means all known and handled exceptions throw the same base exception class and can be handled more easily on a higher level (in modulo components). I'm doing this in C++ and Python.

Two slight changes in the code logic:
- Catch the obvious exception that can happen when translating a joint state message of a certain size to a state rep JointState of different size
- put the clproto encoding and decoding into a try catch block as well and throw a MessageTranslationException in the message translators


eeberhard (approved):
Looks great. It definitely is the right pattern to catch and wrap exceptions at each level. State rep exceptions are caught in modulo core and escalated as CoreException, then modulo_components can catch and escalate as ComponentException where needed. Thanks for the additional throw fields in the docstrings too.

PR #96, opened 2022-07-05
Improve parameter translators for NOT_SET cases

domire8:
Let's do the changes required to fix empty parameter declaration and handling in small steps. Here is the first one with the changes in modulo_core

### DIFF
--- a/source/modulo_core/modulo_core/translators/parameter_translators.py
+++ b/source/modulo_core/modulo_core/translators/parameter_translators.py
@@ -19,7 +19,6 @@ def write_parameter(parameter: sr.Parameter) -> Parameter:
     if parameter.is_empty():
         return Parameter(parameter.get_name(), value=None, type_=Parameter.Type.NOT_SET)
     elif parameter.get_parameter_type() == sr.ParameterType.BOOL or \
-            parameter.get_parameter_type() == sr.ParameterType.BOOL_ARRAY or \
             parameter.get_parameter_type() == sr.ParameterType.INT or \
             parameter.get_parameter_type() == sr.ParameterType.DOUBLE or \
             parameter.get_parameter_type() == sr.ParameterType.STRING:


domire8:
Even though we realized we have to overhaul a few things with the parameters, this PR is still important. It currently fails because it depends on a PR in control libraries:

https://github.com/epfl-lasa/control-libraries/pull/301

source/modulo_core/modulo_core/translators/parameter_translators.py
```diff
@@ -15,7 +16,10 @@ def write_parameter(parameter: sr.Parameter) -> Parameter:
     :raises ParameterTranslationError if the parameter could not be written
     :return: The resulting ROS parameter
     """
-    if parameter.get_parameter_type() == sr.ParameterType.BOOL or \
+    if parameter.is_empty():
+        return Parameter(parameter.get_name(), value=None, type_=Parameter.Type.NOT_SET)
+    elif parameter.get_parameter_type() == sr.ParameterType.BOOL or \
+            parameter.get_parameter_type() == sr.ParameterType.BOOL_ARRAY or \
```
buschbapti, line 22: This one is weird as it appears below as well

buschbapti (dismissed):

source/modulo_core/src/translators/parameter_translators.cpp
```diff
@@ -184,15 +188,18 @@ std::shared_ptr<state_representation::ParameterInterface> read_parameter(const r
...
+std::shared_ptr<ParameterInterface> read_parameter_const(
+    const rclcpp::Parameter& ros_parameter, const std::shared_ptr<const ParameterInterface>& parameter
 ) {
   if (ros_parameter.get_name() != parameter->get_name()) {
     throw exceptions::ParameterTranslationException(
         "The ROS parameter " + ros_parameter.get_name()
             + " to be read does not have the same name as the reference parameter " + parameter->get_name());
   }
+  if (ros_parameter.get_type() == rclcpp::PARAMETER_NOT_SET) {
+    return make_shared_parameter_interface(
```
eeberhard, line 200: This function was a nice way to resolve the empty copy / construction for NOT SET 👍

eeberhard (dismissed):
Seems good once it builds, except for the BOOL_ARRAY duplication that @buschbapti caught in parameter_translators.py

PR #97, opened 2022-07-06
Handle empty parameters in ComponentInterface

domire8:
Okay now I think we're getting somewhere..There are a few changes to the logic here but most of the lines come from the tests

- Add parameter now throws an exception if the parameter could not be added for some reason (translation failed or validation failed). Note that upon declaration of a `NOT_SET` parameter in C++, the set parameter callback is not called, hence we need to validate that parameter explicitly in the `add_parameter`.
- Even though, empty parameters are declared with `dynamic_typing=true`, it's not possible to change the parameter type once the parameter was set because `read_parameter_const` will throw an exception and the callback will return false.
- Along the other changes, I added the `read_only` field for python ROS parameters and it works as in C++ (tested)

Also another few key points that came up and that we have to keep in mind:
- I'd be glad if we would only hide the ROS methods in the Component and LifecycleComponent and not in ComponentInterface, such that we can still do all those tests
- We are calling a virtual method `validate_parameter` from the constructor, where we also add the `period` parameter. This means that the period parameter is still validated with the virtual method that just returns true because the base class has not been fully constructed yet. Only for parameters added in the constructor of any derived component, the override `validate_parameter` will be called. We could check the period for meaningful values, e.g. between 1ms and 10s or something and issue a warning otherwise.

### DIFF
--- a/source/modulo_components/CMakeLists.txt
+++ b/source/modulo_components/CMakeLists.txt
@@ -21,7 +21,7 @@ find_package(ament_cmake_python REQUIRED)
 
 ament_auto_find_build_dependencies()
 
-find_package(control_libraries 6.0.2 REQUIRED COMPONENTS state_representation)
+find_package(control_libraries 6.0.6 REQUIRED COMPONENTS state_representation)
 find_package(clproto 6.0.0 REQUIRED)
 
 include_directories(include)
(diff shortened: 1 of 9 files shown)


domire8:
This seems to work as desired now, at least with all the base components that I've already migrated

eeberhard (approved):
After the back and forth, I believe this is the right approach 👍

PR #98, opened 2022-07-06
Don't start execute thread on construction

domire8:
To be fair, we should even think about removing that option entirely (especially in C++), because we could not think of a component that would want to start a thread on execution and I think it's even really dangerous because we are calling a virtual function from the constructor. In Python it's not such a big problem


eeberhard:
After our discussions, I would definitely prefer to remove this constructor argument entirely

buschbapti (approved):

eeberhard (approved):

PR #99, opened 2022-07-07
Improved exception handling

domire8:
Lots of line changes but mostly from putting try catch blocks around a big block of code.

I had to dig into those exceptions because I got a warning that an exception was not handled correctly when I was trying to use a new component and the most crucial fix was the line where I only handled the TransformError in the component_interface.py and not the MessageTranslationError from the line before.

But while I was there, I improved the exception handling in a few other places too, namely in the subscription callback from the SubscriberInterface and the publish method of the PublisherInterface

### DIFF
--- a/source/modulo_core/src/communication/PublisherInterface.cpp
+++ b/source/modulo_core/src/communication/PublisherInterface.cpp
@@ -106,7 +106,7 @@ void PublisherInterface::publish() {
         break;
     }
   } catch (const exceptions::CoreException& ex) {
-    throw exceptions::CoreException(ex);
+    throw;
   }
 }
 


domire8:
This PR has lower priority than the other ones currently open

eeberhard (dismissed):
Looks good! Only a very minor comment

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -318,8 +318,9 @@ def __subscription_callback(self, message: MsgT, attribute_name: str, reader: Ca
         """
         try:
             self.__setattr__(attribute_name, reader(message))
-        except Exception as e:
-            self.get_logger().error(f"Failed to read message for attribute {attribute_name}", throttle_duration_sec=1.0)
+        except (AttributeError, MessageTranslationError) as e:
+            self.get_logger().warn(f"Failed to read message for attribute {attribute_name}: {e}",
```
buschbapti, line 322: I don't think this should be a `warn` but an `error` as the others.

source/modulo_core/src/communication/SubscriptionHandler.cpp
```diff
@@ -6,73 +6,103 @@ namespace modulo_core::communication {
 
 template<>
 SubscriptionHandler<std_msgs::msg::Bool>::SubscriptionHandler(std::shared_ptr<MessagePairInterface> message_pair) :
-    SubscriptionInterface(std::move(message_pair)) {}
+    SubscriptionInterface(std::move(message_pair)), clock_(std::make_shared<rclcpp::Clock>()) {}
```
buschbapti, line 9: I am just seeing that now but is there a reason you are creating a new clock and not passing the one from the parent node at construction?

buschbapti (commented):
Some comments but looks good otherwise

source/modulo_components/modulo_components/component_interface.py
eeberhard: I think warn is fine here. We will have to write the general guidelines for this and other things, but I would suggest we leave `error` for cases where something fails and leave the component in a different state than expected. For example,
- a user-requested transition callback or service fails
- an run-time exception cannot be handled and `raise_error` is called

I would suggest warn is more appropriate for periodic things:
- subscription callback cannot read the type
- on step callback is interrupted (but the exception is handled and suppressed)

The difference with `warn` is that the component will still be running and in the same state as before, but we want to log the warning that something is not right.

source/modulo_core/src/communication/SubscriptionHandler.cpp
eeberhard: It doesn't look like the clock object is too expensive to instantiate, and they refer to the same underlying time source. It might even be safer to have multiple clocks rather than one master instance, so that one clock doesn't have to deal with all the callbacks in the queue. However, if we change the clock type away from the default RCL_SYSTEM_TIME at any point, it would be difficult to propagate consistently, but that would only matter for time-critical behaviours. Throttled logging is not time-critical imo. 

TL;DR I think this is fine like this for now, and it makes the construction of the Handler in the component is less verbose

eeberhard (approved):

PR #101, opened 2022-07-09
Minor improvements

domire8:
Just a few small improvements here:
- Link libraries in a more precise way
- Add exceptions and utilities in the __init__.py such that the import of those are a bit simpler
- Upon creation of a signal, a parameter is created for the signal topic. If that parameter already existed, nothing happened. I propose that if the parameter already exists, but is empty, we fill it with the default topic name.

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -600,10 +600,14 @@ inline void ComponentInterface<NodeT>::add_input(
       throw exceptions::AddSignalException("Failed to add input '" + signal_name + "': Input already exists");
     }
     std::string topic_name = default_topic.empty() ? "~/" + parsed_signal_name : default_topic;
-    this->add_parameter(
-        parsed_signal_name + "_topic", topic_name, "Output topic name of signal '" + parsed_signal_name + "'",
-        fixed_topic
-    );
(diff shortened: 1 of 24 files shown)


domire8:
2 more things:
- use the new operator from control libraries 6.0.7
- switch default_topic and read_only arguments in add signal because default topic is most probably used more often

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -600,10 +600,14 @@ inline void ComponentInterface<NodeT>::add_input(
       throw exceptions::AddSignalException("Failed to add input '" + signal_name + "': Input already exists");
     }
     std::string topic_name = default_topic.empty() ? "~/" + parsed_signal_name : default_topic;
-    this->add_parameter(
-        parsed_signal_name + "_topic", topic_name, "Output topic name of signal '" + parsed_signal_name + "'",
-        fixed_topic
-    );
+    auto parameter_name = parsed_signal_name + "_topic";
```
eeberhard, line 603: This addition makes sense. It also doesn't counteract the discussion we had about adding and removing signals - since you are already checking for the existence of the parameter it will be easier to integrate that planned feature in a future PR

eeberhard (approved):
Looks good to me 👍

PR #102, opened 2022-07-14
Add trigger and send predicate on set

domire8:
Two minor features that we discussed over the past days:
- Send a predicate value immediately after its set, on top of the periodic publishing
- Add a trigger that is essentially a predicate that is always false except when its triggered it will be evaluated to true exactly once.

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -611,6 +611,7 @@ inline void ComponentInterface<NodeT>::trigger(const std::string& trigger_name) 
     return;
   }
   this->triggers_.at(trigger_name) = true;
+  publish_predicate(trigger_name);
 }
 
 template<class NodeT>
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -166,6 +166,8 @@ protected:
 
   /**
    * @brief Set the value of the predicate given as parameter, if the predicate is not found does not do anything.
+   * @details Even though the predicates are published periodically, the new value of this predicate will be published
+   * once immediately after setting it.
    * @param predicate_name the name of the predicate to retrieve from the map of predicates
    * @param predicate_value the new value of the predicate
    */
@@ -173,6 +175,8 @@ protected:
 
   /**
    * @brief Set the value of the predicate given as parameter, if the predicate is not found does not do anything.
+   * @details Even though the predicates are published periodically, the new value of this predicate will be published
+   * once immediately after setting it.
    * @param predicate_name the name of the predicate to retrieve from the map of predicates
    * @param predicate_function the function to call that returns the value of the predicate
    */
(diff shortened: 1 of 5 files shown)


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -557,6 +581,38 @@ inline bool ComponentInterface<NodeT>::get_predicate(const std::string& predicat
...
+}
+
+template<class NodeT>
+inline void ComponentInterface<NodeT>::trigger(const std::string& trigger_name) {
+  if (this->triggers_.find(trigger_name) == this->triggers_.end()) {
+    RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to trigger: could not find trigger"
+                                            " with name  '" << trigger_name << "'.");
+    return;
+  }
+  this->triggers_.at(trigger_name) = true;
```
eeberhard, line 613: I think this would warrant triggering the publication, as it is in `set_predicate`
```suggestion
  this->triggers_.at(trigger_name) = true;
  publish_predicate(trigger_name);
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -266,6 +269,46 @@ def set_predicate(self, name: str, value: Union[bool, Callable[[], bool]]):
...
+    def trigger(self, trigger_name: str):
+        """
+        Latch the trigger with the provided name.
+
+        :param trigger_name: The name of the trigger
+        """
+        if not trigger_name in self._triggers.keys():
+            self.get_logger().error(f"Failed to trigger: could not find trigger with name '{trigger_name}'.")
+            return
+        self._triggers[trigger_name] = True
```
eeberhard, line 311:
```suggestion
        self._triggers[trigger_name] = True
        self._publish_predicate(trigger_name)
```

source/modulo_components/test/cpp/test_component_interface.cpp
```diff
@@ -139,4 +139,18 @@ TYPED_TEST(ComponentInterfaceTest, RaiseError) {
...
+  ASSERT_FALSE(this->component_->triggers_.find("trigger") == this->component_->triggers_.end());
+  EXPECT_FALSE(this->component_->triggers_.at("trigger"));
+  EXPECT_FALSE(this->component_->get_predicate("trigger"));
+  EXPECT_NO_THROW(this->component_->trigger("trigger"));
+  // After triggering, the trigger will be true only once
+  EXPECT_TRUE(this->component_->triggers_.at("trigger"));
+  EXPECT_TRUE(this->component_->get_predicate("trigger"));
+  // After the predicate function was evaluated once, the trigger is back to false
+  EXPECT_FALSE(this->component_->triggers_.at("trigger"));
+  EXPECT_FALSE(this->component_->get_predicate("trigger"));
```
eeberhard, line 154: My suggestion about publishing the predicate on triggering will break this test case. Without adding a predicate listener test, you could change it to:

```suggestion
  // When reading, the trigger will be true only once
  this->component_->triggers_.at("trigger") = true;
  EXPECT_TRUE(this->component_->triggers_.at("trigger"));
  EXPECT_TRUE(this->component_->get_predicate("trigger"));
  // After the predicate function was evaluated once, the trigger is back to false
  EXPECT_FALSE(this->component_->triggers_.at("trigger"));
  EXPECT_FALSE(this->component_->get_predicate("trigger"));
```

source/modulo_components/test/python/test_component_interface.py
```diff
@@ -78,3 +78,17 @@ def test_tf(component_interface):
...
     assert abs(identity.get_orientation().w) - 1 < 1e-3
+
+
+def test_add_trigger(component_interface):
+    component_interface.add_trigger("trigger")
+    assert "trigger" in component_interface._triggers.keys()
+    assert not component_interface._triggers["trigger"]
+    assert not component_interface.get_predicate("trigger")
+    component_interface.trigger("trigger")
+    # After triggering, the trigger will be true only once
```
eeberhard, line 89: same comment here

eeberhard (commented):
It's a useful feature. Just some details

eeberhard (approved):

PR #103, opened 2022-07-15
Component services

eeberhard:
For components to have simple services that can trigger certain behaviours, it is necessary to define some service messages. A new package defines these component interfaces.

The new component add_service methods allow simple creation of these services with registered callbacks, either taking no arguments or a single string argument. They return a status and optional message. This should allow flexible and general usage while limiting the number of specific service types that need to be supported.

Still some TODOs for documentation and better tests but I think the general idea is ready for review

### DIFF
--- a/source/modulo_components/modulo_components/component_interface.py
+++ b/source/modulo_components/modulo_components/component_interface.py
@@ -423,7 +423,7 @@ class ComponentInterface(Node):
         except Exception as e:
             self.get_logger().error(f"Failed to add input '{signal_name}': {e}")
 
-    def add_service(self, service_name: str, callback: Union[Callable[[], dict], Callable[[string], dict]]):
+    def add_service(self, service_name: str, callback: Union[Callable[[], dict], Callable[[str], dict]]):
         """
         Add a service to trigger a callback function.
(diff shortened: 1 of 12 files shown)


domire8 (approved):
Nice!!

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -421,6 +423,61 @@ def add_input(self, signal_name: str, subscription: Union[str, Callable], messag
...
+                        # if the dict does not contain a success field, but the callback
+                        # completes without error, assume it was successful
+
+                        response.success = True
+                    if "message" in ret.keys():
+                        response.message = ret["message"]
+                elif isinstance(ret, bool):
+                    response.success = ret
+                else:
+                    response.success = True
```
eeberhard, line 461: This can be simplified a little

```suggestion

                # if the return does not contain a success field or bool result,
                # but the callback completes without error, assume it was successful
                response.success = True
                if isinstance(ret, dict):
                    if "success" in ret.keys():
                        response.success = ret["success"]
                    if "message" in ret.keys():
                        response.message = ret["message"]
                elif isinstance(ret, bool):
                    response.success = ret
```

PR #104, opened 2022-07-18
Create service under local namespace

eeberhard:
* Formatting fixes in log lines

I missed something in #103 - though I tested that the services were created, I forgot about the namespace. The intended syntax is that some `service_endpoint` on component `foo` becomes available on the topic `/foo/service_endpoint`, so that the AICA state engine ComponentInterface can automatically associate the service endpoint with the target component. While add_input provides a "default topic" option for flexible renaming or remapping, I believe services should be more static in their registered namespace. Signals can be renamed to connect components together, but services are always endpoints associated with a component, and so should be in the namespace of that component.

This has been tested and works as intended.

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -794,9 +794,9 @@ inline void ComponentInterface<NodeT>::add_service(
 ) {
   try {
     std::string parsed_service_name = utilities::parse_signal_name(service_name);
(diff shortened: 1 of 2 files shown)


eeberhard:
I know it's not standard protocol, but given that the main reviewers are on holiday or otherwise busy, and that this PR is a small amendment to a new feature with no other downstream users yet, I will go ahead and merge it so that I can continue developing without having to propagate a custom-branch CI image

PR #105, opened 2022-07-19
Add service improvements

eeberhard:
Here are some general improvements to the recent add_service feature to increase safety and reduce code duplications. I mirrored some of the improvements in the `add_input` function since the behaviour is similar.

Review by commit for more readability.


buschbapti (approved):
Very nice solution to remove code duplication.

PR #106, opened 2022-07-21
Generate documentation

eeberhard:
I am revising and extending documentation in anticipation of the 2.0 release. As part of this, I want to easily serve the existing source code documentation which we have already provided in the doxygen format.

This workflow and structure is equivalent to the one I developed for control libraries. A GitHub workflow generates html documentation based on a doxygen.conf file. It moves that html directory to a named subdirectory (e.g. `versions/develop`) and pushes a commit to the dedicated `docs` branch. GitHub Pages is configured to serve the root of `docs`. Here, an index page links to all documented versions and branches in the `versions` subdirectory. 

I already manually generated the docs for the current releases (v1.0.0 and v1.1.0), main and develop. You can view them here: https://epfl-lasa.github.io/modulo


domire8 (approved):
Neat, thanks!

PR #108, opened 2022-08-02
Improve State message reading

domire8:
This is something I started before going on holidays and should go in before the release. Reading shared pointers of State with incompatible types were leading to segfaults, they are now being caught and an exception is raised. Additionally, where applicable with JointState and CartesianState types, messages of the derived type can also be read as the parent type (the commit message is wrong there).

I'd be glad if you could help me write a concise function header and maybe find a better function name for `dynamic_cast_and_check` because I'd rather have something precise than an unnecessary and incorrect doc.


eeberhard (approved):
This was a bigger PR than expected, but it's a very important part of the framework so I appreciate the effort. The message encoding tests are very thorough, good job on those 👍

PR #110, opened 2022-08-04
Improved message translators suggestions

eeberhard:
I was reviewing PR #108 and was adding the documentation as requested, when I thought of a slightly better pattern for all the value conversions. I opened this as a separate PR so that you can decide if you like it. The rest of PR #108 is otherwise very good in my opinion.


eeberhard:
You may have also seen I used `std::set` for the spatial state supported types. Range-based comparison of enums is brittle and should generally be avoided; you can't assume that the literal value and order of each enum won't change at a future point, and if they do it's the kind of downstream issue that's really hard to predict. (that said StateType will rarely change because the exact enumeration order is tied up with in the clproto encoding)

PR #111, opened 2022-08-05
Catch core and component exceptions

domire8:
There were some smaller TODOs left regarding exception handling that I tried to resolve here. With the publisher handler (hopefully) only throwing CoreExceptions, we can just catch those explicitely.

### DIFF
--- a/source/modulo_components/include/modulo_components/Component.hpp
+++ b/source/modulo_components/include/modulo_components/Component.hpp
@@ -144,9 +144,7 @@ inline void Component::add_output(
         break;
       }
     }
-  } catch (const modulo_core::exceptions::CoreException& ex) {
-    RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to add output '" << signal_name << "': " << ex.what());
-  } catch (const exceptions::ComponentException& ex) {
+  } catch (const std::exception& ex) {
     RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to add output '" << signal_name << "': " << ex.what());
   }
 }
--- a/source/modulo_components/include/modulo_components/Component.hpp
+++ b/source/modulo_components/include/modulo_components/Component.hpp
@@ -7,6 +7,7 @@
 #include "modulo_components/ComponentInterface.hpp"
 #include "modulo_components/utilities/utilities.hpp"
 #include "modulo_core/EncodedState.hpp"
+#include "modulo_core/exceptions/CoreException.hpp"
 
 namespace modulo_components {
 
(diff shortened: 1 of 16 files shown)


source/modulo_components/include/modulo_components/Component.hpp
```diff
@@ -143,8 +144,9 @@ inline void Component::add_output(
         break;
       }
     }
-  } catch (const std::exception& ex) {
-    // TODO if modulo::communication had a base exception, could catch that
+  } catch (const modulo_core::exceptions::CoreException& ex) {
+    RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to add output '" << signal_name << "': " << ex.what());
```
eeberhard, line 148: In these cases when there are identical branches, it discards the benefit of catching the custom exceptions. I would suggest just adding that additional info to the logger (since `ex.what()` does not actually tell us the exception type). This kind of thing might make debugging from logs easier in future.
```suggestion
    RCLCPP_ERROR_STREAM(this->get_logger(), "Failed to add output '" << signal_name << "'! CoreException: " << ex.what());
```

source/modulo_core/CMakeLists.txt
```diff
@@ -21,8 +21,8 @@ find_package(ament_cmake_python REQUIRED)
 
 ament_auto_find_build_dependencies()
 
-find_package(control_libraries 6.0.7 REQUIRED COMPONENTS state_representation)
-find_package(clproto 6.0.0 REQUIRED)
+find_package(control_libraries 6.0.8 REQUIRED COMPONENTS state_representation)
```
eeberhard, line 24: Since you already changed the required CL version here you might as well update it to 6.1 now following the latest release!

eeberhard (commented):
This PR gave me an idea. Since we are catching base exceptions and ultimately just logging the exception message, it might be nice to know which derived exception class is responsible. This would require some additional introspection since the contents of `exception::what()` does not tell us what the exception class actually is. But, we could actually embed that information into our exception classes with a predefined format!

```c++
class CoreException : public std::runtime_error {
public:
  explicit CoreException(const std::string& msg) : CoreException("CoreException", msg) {};
protected:
  CoreException(const std::string& msg, const std::string& prefix) : std::runtime_error(prefix + ": " + msg) {};
};
```

Then for a derived exception class we could just do:
```c++
class NullPointerException : public CoreException {
public:
  explicit NullPointerException(const std::string& msg) : CoreException("NullPointerException", msg) {};
};
```

Let me know what you think. This PR is small enough (when ignoring whitespace changes) that I think it might be a nice change to try

domire8:
> Let me know what you think. This PR is small enough (when ignoring whitespace changes) that I think it might be a nice change to try

I do like this because I was always wondering for myself a bit why we we give exceptions a verbose and descriptive name if the user never sees that. So yeah I can add that here :+1:

eeberhard (commented):
Nice 👍

eeberhard (approved):

PR #112, opened 2022-08-05
Modulo components documentation

eeberhard:
Here's the last documentation push. I think some concepts can still be described in more detail and the CONTRIBUTING doc is still a bit incomplete, but in general I would probably leave further details (concepts and usage guidelines) to the Component SDK next to tangible examples


domire8 (approved):
Beautiful, thanks a lot for that, much appreciated

PR #117, opened 2022-08-05
Support multiple build test actions

eeberhard:
* Move build-test entrypoint script to a common directory

* Remove local Dockerfile from action and instead directly reference the base ghcr.io image and common build-test entrypoint script

* Add build test action variant for galactic devel image

* Invoke build test action job for galactic devel

---

Modulo is built on top of the [ghcr.io/aica-technology/ros2-control-libraries](https://github.com/aica-technology/docker-images/pkgs/container/ros2-control-libraries) image, which has some variant base tags, namely `galactic` and `galactic-level`. These are built with the `main` and `develop` branches of [Control Libraries](https://github.com/epfl-lasa/control-libraries) respectively.

When developing features for Modulo, it would be good to test the compatibility of both base tags. More specifically, it would inform us when a new Modulo release is blocked until the next Control Libraries release (e.g. when the modulo `develop` branch uses features that only exist on CL `develop` and not CL `main`). For this we can extend the build-test action and workflow.

Because of the limitations of [GitHub Docker Actions](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runs-for-docker-container-actions), there was no good way to set the base image of the Dockerfile through an ARG or input dynamically. So it seems we need two separate actions, one for each base tag. But the build-test entrypoint.sh script contains fair amount of test logic, and I didn't want to have to duplicate that file. The Dockerfile itself can also not COPY any common build-test script from an upper directory, because it can only COPY files from the docker build root (which is the `.github/actions/<action>` directory).

Eventually I realised that I could access a common build-test script by referring to the `/github/workspace` directory that is automatically copied into the image. Furthermore, I could refer to a hosted image and override the entrypoint directly in the action file, eliminating the need for a separate local Dockerfile. With all of this in place, it is very simple to add new actions for different base images without needing to duplicate much code at all.

Regarding branch protection rules, I would enforce that the `Galactic development build and test` job must pass for PRs onto `develop`, and that the `Galactic build and test` job must pass for PRs onto `main`.


domire8:
Looks nice, what happens in the PRs to develop if the galactic tests don't pass? We'll only have 3/4 tests that pass, resulting in an overall test failure indicated with a red cross no?

eeberhard:
>We'll only have 3/4 tests that pass

Yes, that would indeed be the outcome. While the ❌ is not particularly nice to look at, the PR can still be valid because the `galactic` check is not required for PRs into `develop`. I think this is a fair compromise because it's also not a situation I expect will occur frequently. Most of the time we will be using control libraries features that are already released, and so 4/4 checks will pass ✅ . At some point in development we will hit a point where some new CL feature is needed, and there will be a streak of PRs that pass 3/4, but that will prompt us to make the minor release in CL. It's sort of a way of keeping us accountable and releasing more regularly.

domire8 (approved):

PR #118, opened 2022-08-10
Manually export modulo_core dependencies

domire8:
By using the ament directives manually instead of using ament_auto, it is possible to export the modulo_core dependencies (especially clproto) correctly such that downstream packages don't have to repeat finding and linking those anymore.


eeberhard (commented):
Thanks for taking the time to revise this. It looks good as far as I can tell.

The package.xml still has the auto dependencies - are they both still needed? In any case I don't think it does any harm to keep them there, just curious.
```xml
<buildtool_depend>ament_cmake_auto</buildtool_depend>
<buildtool_depend>ament_cmake_python</buildtool_depend>
```

Have you been able to test it with a downstream package? If it resolves the mentioned issue then it's definitely good to go 👍

domire8:
> Have you been able to test it with a downstream package? If it resolves the mentioned issue then it's definitely good to go +1

Yes tested together with the other PR on modulo-controllers and on the robot :+1:

eeberhard (approved):

PR #119, opened 2022-08-11
Ensure compatibility with humble

domire8:
A few more updates in the CMakeLists and modulo build with humble. It seems humble is more precise with all those ament directives. They are obviously also correct like that for galactic


buschbapti (dismissed):

source/modulo_core/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ find_package(ament_cmake_python REQUIRED)
 find_package(geometry_msgs REQUIRED)
 find_package(sensor_msgs REQUIRED)
 find_package(std_msgs REQUIRED)
-find_package(rclcpp REQUIRED)
+find_package(rclcpp_lifecycle REQUIRED)
```
eeberhard, line 25: 🤨 I see why adding `rclcpp_lifecycle` dependency is a good idea, but I don't like the removal of `rclcpp`. I would prefer to have both at least (and not the implicitly take rclcpp as a dependency of lifecycle) - same for target / export dependencies

domire8: You are totally right there

eeberhard (approved):

domire8:
Forgot that the tests here are not representative for humble so I was able to catch another thing early: We have to rename `_services` in the component_interface.py to something else because it will override an attribute from the rclpy node otherwise

PR #120, opened 2022-08-15
Add static tf broadcaster

eeberhard:
* Add static tf broadcaster as class property, with methods
to add static broadcaster and send static tf

* In C++, pass default publisher options to static transform
broadcaster constructor, instead of the default-constructed
qos override options

* Add helper function to publish the transform in a generic way
to support static and non-static broadcasters, and thereby
reduce code duplication in the user-facing send transforms

* Add tests


domire8 (approved):
Nice addition :+1:

buschbapti (approved):

PR #121, opened 2022-08-16
Use prefix for all exceptions

domire8:
Something that we started in modulo_core C++ exceptions was to use a prefix in the exception messages to have a better idea of where the exception is coming from. I propagated that idea to python and to modulo_components


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -738,12 +738,10 @@ template<class NodeT>
...
   std::string parsed_signal_name = utilities::parse_topic_name(signal_name);
   if (parsed_signal_name.empty()) {
-    throw exceptions::AddSignalException(
-        "Failed to add input '" + signal_name + "': Parsed signal name is empty."
-    );
+    throw exceptions::AddSignalException("Parsed signal name is empty.");
   }
   if (this->inputs_.find(parsed_signal_name) != this->inputs_.cend()) {
-    throw exceptions::AddSignalException("Failed to add input '" + signal_name + "': Input already exists");
+    throw exceptions::AddSignalException("Input already exists");
```
eeberhard, line 744: You don't think it's worth keeping the `signal_name` in the error message?

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -852,13 +850,11 @@ template<class NodeT>
...
   if (parsed_service_name.empty()) {
-    throw exceptions::AddServiceException(
-        "Failed to add service '" + service_name + "': Parsed service name is empty."
-    );
+    throw exceptions::AddServiceException("Parsed service name is empty.");
   }
   if (this->empty_services_.find(parsed_service_name) != this->empty_services_.cend()
       || this->string_services_.find(parsed_service_name) != this->string_services_.cend()) {
-    throw exceptions::AddServiceException("Failed to add service '" + service_name + "': Service already exists");
+    throw exceptions::AddServiceException("Service already exists");
```
eeberhard, line 857: Same point for `service_name`

eeberhard (commented):
I appreciate the change, but I do think it's worth preserving the name information in the signal and service cases. Not in the empty parsed name though, that's fine to remove

source/modulo_components/include/modulo_components/ComponentInterface.hpp
domire8: I dont have it in front of me but if I remember correctly from yesterday, this exception is caught a few lines further down and the log message contains the signal name. So if I dont remove that, you end up with messages like "Failed to add input 'test': AddSignalException: Failed to add input 'test'"

domire8:
If you run the tests in CLion you will see that the log messages you get still contain all the infos we are intersted in. I just removed duplicated info

source/modulo_components/include/modulo_components/ComponentInterface.hpp
eeberhard: Thanks for the clarification, that was easy to miss just looking at this diff - in that case I definitely appreciate the reduction in code duplication!

eeberhard (approved):

PR #122, opened 2022-08-16
Accept list of transforms in send_transform

buschbapti:
Following ROS API, this PR adds the possibility to use `send_transform` with a list of transform and not just a single transform. For that, the behavior is that the helper function `publish_transform` accept by default a list of transforms and when a single one is provided in `send_transform` it is expanded to a list.

This passes the tests both in python and c++, although I haven't yet made a specific test for the lists, which I would do in a next commit.


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -371,15 +383,15 @@ class ComponentInterface : public NodeT {
   void evaluate_periodic_callbacks();
 
   /**
-   * @brief Helper function to send a transform through a transform broadcaster
+   * @brief Helper function to send a vector of transform through a transform broadcaster
```
eeberhard, line 386:
```suggestion
   * @brief Helper function to send a vector of transforms through a transform broadcaster
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -628,11 +631,11 @@ def _evaluate_periodic_callbacks(self):
...
                                         throttle_duration_sec=1.0)
 
-    def __publish_transform(self, transform: sr.CartesianPose, static: Bool=False):
+    def __publish_transform(self, transforms: List[sr.CartesianPose], static: Bool=False):
         """
-        Send a transform to TF using the normal or static tf broadcaster
+        Send a list of transforms to TF using the normal or static tf broadcaster
 
-        :param transform: The transform to send
+        :param transforms: The list of transform to send
```
eeberhard, line 638:
```suggestion
        :param transforms: The list of transforms to send
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -2,6 +2,7 @@
 import sys
 from functools import partial
 from typing import Callable, Dict, List, Optional, TypeVar, Union
+from collections import Iterable
```
eeberhard, line 5:
```suggestion
from collections.abc import Iterable
```

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -512,21 +513,23 @@ def add_tf_listener(self):
...
         """
-        Send a static transform to TF.
+        Send a static transform or a list of static transforms to TF.
 
-        :param transform: The transform to send
+        :param transform:  The transform or list of transforms to send
         """
-        self.__publish_transform(transform, static=True)
+        list_transforms = transform if isinstance(transform, Iterable) else [transform]
+        self.__publish_transform(list_transforms, static=True)
```
eeberhard, line 532: If the type of `transform` is either `sr.CartesianPose` or `Iterable[sr.CartesianPose]`, then you can just do
```suggestion
        self.__publish_transform(list(transform))
```
The `list` constructor will iterate over the object to create a list, even if the iterable is already a list

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -328,12 +328,24 @@ class ComponentInterface : public NodeT {
    */
   void send_transform(const state_representation::CartesianPose& transform);
 
+  /**
+   * @brief Send a vector of transforms to TF.
+   * @param transforms The vector of transform to send
```
eeberhard, line 333:
```suggestion
   * @param transforms The vector of transforms to send
```

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -328,12 +328,24 @@ class ComponentInterface : public NodeT {
...
+
   /**
    * @brief Send a static transform to TF.
    * @param transform The transform to send
    */
   void send_static_transform(const state_representation::CartesianPose& transform);
 
+  /**
+   * @brief Send a vector of static transforms to TF.
+   * @param transforms The vector of transform to send
```
eeberhard, line 345:
```suggestion
   * @param transforms The vector of transforms to send
```

source/modulo_components/modulo_components/component_interface.py
eeberhard: Or preferably, 
```python3
from typing import Callable, Dict, List, Optional, TypeVar, Union, Iterable
```

source/modulo_components/modulo_components/component_interface.py
eeberhard: This was wrong, and I was wrong, and I am sorry. The list constructor takes iterables, yes, but _only_ iterables, so it fails for a single transform. The `isinstance` check is also nicer because it doesn't need to iterate and copy construct, so it doesn't get slower for bigger lists.

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -1,7 +1,7 @@
 import inspect
 import sys
 from functools import partial
-from typing import Callable, Dict, List, Optional, TypeVar, Union
+from typing import Callable, Dict, List, Optional, TypeVar, Union, Iterable
```
domire8, line 4: Why do you use Iterable and not List?

domire8 (commented):
I don't like it too much that we use `send_transform` in singular form but it's okay if its ROS terminology. We could at least use `publish_transforms` because there, we really have a for loop and the first argument is already called `transforms

source/modulo_components/modulo_components/component_interface.py
buschbapti: Iterable is more generic than list. It can accept Tuple, Set and such. As long as the container can be iterated through.

buschbapti:
I agree with you and also mentioned that to @eeberhard. Two solutions here, either we follow blindly ROS API or we make two separate functions. The first make it easier to remember which function to use, although it would be quite explicit. The later would make the python version a bit nicer as you would not need to test for the input type. I don't mind either to be honest.

domire8:
> I don't mind either to be honest.

Overloaded `send_transform` I can live with although I'd prefer `send_transforms`. But for the `publish_transform` helper I'd really prefer the plural form with 's'

buschbapti:
Test have now been added. This is good to go after final check.

domire8 (approved):

buschbapti:
> Overloaded send_transform I can live with although I'd prefer send_transforms. But for the publish_transform helper I'd really prefer the plural form with 's'

I also have a preference for `send_transforms` to be honest. We are already breaking ros design by having a `send_static_transform` function so I would suggest to break it further but being more explicit.

buschbapti:
Las commit changes to `send_transforms` and `send_static_transforms`. If @eeberhard prefers the overloaded version we can always revert it.

eeberhard (approved):
Separate functions are fine 👍 I don't care to mirror the ROS API, I just want the most convenience, and this way gives the most clarity

domire8 (approved):

buschbapti:
Perfect then I will merge that now.

PR #124, opened 2022-08-24
User-defined callback on inputs (C++)

eeberhard:
[Insert a user callback into SubscriptionHandler](https://github.com/epfl-lasa/modulo/commit/b64a7e0bee1a1ab320770dd4da38ed11a787299c)
* Add private class property and setter for user_callback_, a void function

* Invoke user callback within the subscription callback, after reading the message

* Move callback error handling to helper function to simplify implementation and reduce code duplication

* Allow setting user callback directly when getting the subscription callback with a function overload

* Update test to use user callback

* Adjust docstrings

[Implement user callback in component add_input](https://github.com/epfl-lasa/modulo/commit/c56fef8fc9deaabee1418ec7a52c046ffebe85b7)

* Resolve TODO by overloading add_input with a user_callback argument after the shared data pointer. This allows a void function to be executed after the message is read into the data (as opposed to the generic input callback which takes a raw message). The default add_input with a data pointer passes an empty user callback instead.

* Update tests

---

This was a quick little feature I wanted to implement. Basically it allows extra logic to be triggered on a subscription, but while retaining all the convenience of the standard automatic message reading when providing the data pointer. I chose to inject it into the SubscriptionHandler so that we can guarantee the callback is only triggered when the data has been read successfully. The tests pass but I haven't tested a real component with it.

TODO is the equivalent Python implementation. Maybe @domire8 would be interested to do that.

### DIFF
--- a/source/modulo_components/include/modulo_components/ComponentInterface.hpp
+++ b/source/modulo_components/include/modulo_components/ComponentInterface.hpp
@@ -228,12 +228,25 @@ protected:
    * @param default_topic If set, the default value for the topic name to use
    * @param fixed_topic If true, the topic name of the input signal is fixed
    */
-  // TODO could be nice to add an optional callback here that would be executed from within the subscription callback
-  // in order to manipulate the data pointer upon reception of a message
   template<typename DataT>
   void add_input(
       const std::string& signal_name, const std::shared_ptr<DataT>& data, const std::string& default_topic = "",
       bool fixed_topic = false
+  );
+
+  /**
+   * @brief Add and configure an input signal of the component.
+   * @tparam DataT Type of the data pointer
+   * @param signal_name Name of the input signal
+   * @param data Data to receive on the input signal
+   * @param callback Callback function to trigger after receiving the input signal
+   * @param default_topic If set, the default value for the topic name to use
+   * @param fixed_topic If true, the topic name of the input signal is fixed
+   */
+  template<typename DataT>
+  void add_input(
+      const std::string& signal_name, const std::shared_ptr<DataT>& data, const std::function<void()>& callback,
+      const std::string& default_topic = "", bool fixed_topic = false
   );
 
   /**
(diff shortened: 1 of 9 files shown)


domire8 (commented):
That already look wonderful!

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -894,11 +894,13 @@ template<class NodeT>
 inline std::string ComponentInterface<NodeT>::validate_service_name(const std::string& service_name) {
   std::string parsed_service_name = utilities::parse_topic_name(service_name);
   if (parsed_service_name.empty()) {
-    throw exceptions::AddServiceException("Parsed service name is empty.");
+    throw exceptions::AddSignalException(
```
eeberhard, line 897: You accidentally changes the exception type, it should still be `AddServiceException` here and a few lines below

source/modulo_components/modulo_components/component_interface.py
```diff
@@ -467,15 +468,16 @@ def callback_wrapper(request, response, cb):
         try:
             parsed_service_name = parse_topic_name(service_name)
             if not parsed_service_name:
-                raise AddServiceError(f"Parsed service name is empty.")
+                raise AddSignalError(f"The parsed signal name for service {service_name} is empty. Provide a "
```
eeberhard, line 471: Another copy error, should be `AddServiceError` and `f"The parsed service name ... "`

source/modulo_components/include/modulo_components/ComponentInterface.hpp
domire8: Ouf good catch thanks

domire8:
@eeberhard @buschbapti I propose to do this feature in several stages because it's finally time to add more component tests and the new feature with the additional user callback simplifies this greatly (because we can add a future very easily in the subscription callback).  Hence, this PR includes @eeberhard commits as well as an additional commit containing component level test with minimal input and output nodes, where the new feature is already tested (only in C++ for now).

The base branch is a feature branch too so I'd merge this there and then we can worry about changelog and version number at the end.

eeberhard (commented):
Great suggestion, and thanks for moving this forward.

Since I authored this PR I can't officially "approve", so instead I'll just close it into the feature branch.

Issue #125, opened 2022-08-25
Inconsistent use of underscores for protected functions in Python


domire8:
We have small inconsistencies for the single and double underscore pattern for protected and private functions in Python. For example,  the `_validate_parameter` function in `ComponentInterface` has an underscore to indicate its protected but the `on_XXX_callback` functions in `LifecylceComponent` don't have that.

I'd prefer to remove the single underscores and only mark private functions with double underscore. This is quite a breaking change for the dynamic components but I would take care of it. Any objections @eeberhard @buschbapti ?

domire8:
Ok I guess I'm starting to contradict myself here.

For other functions, such as `_step`, etc. that are defined in the component interface but overridden in the `Component` and `LifecycleComponent` (and hence cannot have double underscore), I'd like to keep the underscore. Maybe in other words:

I'd like to have to following pattern:
- All public functions and functions that should be overridden by a derived component, we use NO underscore.
- All protected functions that are defined in the component interface and required in the `Component` and `LifecycleComponent` we use SINGLE underscore to indicate to the user that those should not be used nor overridden.
- As much as it's possible, we use DOUBLE underscore for all other functions

buschbapti:
Right `step()` is not actually redefined by the final component only the `on_step()`. Then I agree there. Basically, anything publicly usable, even if protected in cpp should be without underscores. To be fair, they could be made public in cpp for consistency but probably that is not necessary.

domire8:
Resolved in #126 and #127

PR #126, opened 2022-08-25
Rename validate parameter callback and validate period value

domire8:
A wrapper for the `validate_parameter` function such that the period parameter can be checked at the Component interface level.


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -589,6 +597,17 @@ inline void ComponentInterface<NodeT>::set_parameter_value(const std::string& na
   }
 }
 
+template<class NodeT>
+inline bool ComponentInterface<NodeT>::validate_parameter_wrapper(
+    const std::shared_ptr<state_representation::ParameterInterface>& parameter
+) {
+  if (parameter->get_name() == "period" && parameter->get_parameter_value<double>() < 1e-3) {
```
eeberhard, line 604: I'm not sure I agree with this magic number limit. 1kHz is certainly a reasonable upper bound but it might change on specialised hardware with real-time determinism. This number should be somewhere higher up so that it can be adjusted if necessary; my suggestion would be a `minimum_period_` private property, for which we can add a protected setter.

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -413,6 +413,14 @@ class ComponentInterface : public NodeT {
    */
   rcl_interfaces::msg::SetParametersResult on_set_parameters_callback(const std::vector<rclcpp::Parameter>& parameters);
 
+  /**
+   * @brief Parameter validation wrapper that validates the period and calls the validation function of the derived
+   * Component classes.
+   * @param parameter A ParameterInterface pointer to a Parameter instance
+   * @return The validation result
+   */
+  virtual bool validate_parameter_wrapper(const std::shared_ptr<state_representation::ParameterInterface>& parameter);
```
eeberhard, line 422: It's marked as `virtual`, but it is also `private`. I think you should remove `virtual`. 
```suggestion
  bool validate_parameter_wrapper(const std::shared_ptr<state_representation::ParameterInterface>& parameter);
```

source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -413,6 +413,14 @@ class ComponentInterface : public NodeT {
    */
   rcl_interfaces::msg::SetParametersResult on_set_parameters_callback(const std::vector<rclcpp::Parameter>& parameters);
 
+  /**
+   * @brief Parameter validation wrapper that validates the period and calls the validation function of the derived
+   * Component classes.
```
eeberhard, line 418: I prefer to put specific behaviours in the details  
```suggestion
   * @brief Parameter validation wrapper
   * @details This validates the period and calls the validation function of the derived Component classes.
```

eeberhard (commented):
I like the concept but not the name `validate_parameter_wrapper`. 

Probably the best and most consistent option would probably be for this to be called `validate_parameter` and for the default overridable validation function to be renamed to `on_validate_parameter` / `on_validate_parameter_callback`. Unfortunately it would require yet another overhaul of existing components.

The non-breaking alternative is to swap that around - have the private wrapper be called `on_validate_parameter` and leave the virtual protected one as `validate_parameter`

A third option, since this is a private function and for now is unique to the base ComponentInterface, is _not_ to act as a wrapper but instead just act as a sequential validation. In that sense, you could call it `validate_internal_parameter` or similar and then do

```c++
if (!this->validate_internal_parameter(new_parameter) || !this->validate_parameter(new_parameter)) {
   ...
}
```

buschbapti:
I like the `on_validate_parameter` change. This makes it clear that every `on_X` are meant to be redefined. I would vote for breaking it again :dagger:

source/modulo_components/include/modulo_components/ComponentInterface.hpp
buschbapti: agree with you. I wouldn't put upper bounds to be honest. Some robots now run at 2kHz

eeberhard:
> I like the `on_validate_parameter` change. This makes it clear that every `on_X` are meant to be redefined. I would vote for breaking it again 🗡️

Especially considering the suggestions for #125 and #127, it would be a good time to refactor the naming for the validation callback. The two options are `on_validate_parameter` and `on_validate_parameter_callback`, but I think the latter is the most consistent.

buschbapti:
> Especially considering the suggestions for #125 and #127, it would be a good time to refactor the naming for the validation callback. The two options are `on_validate_parameter` and `on_validate_parameter_callback`, but I think the latter is the most consistent.

I find `on_validate_parameter_callback` super verbose but I tend to agree for consistency.

source/modulo_components/include/modulo_components/ComponentInterface.hpp
domire8: Copy error of course, thanks

source/modulo_components/include/modulo_components/ComponentInterface.hpp
eeberhard: Actually, thinking about this more, I think it should just be invalid at `<= 0.0`. Any positive real number will work because it signifies the target periodicity - if the executor / CPU can't keep up then the real period will be slower automatically. At some point we'll add watchdog monitoring and diagnostics to the main engine so we can see the process load, in which case we can warn if a component period is too low to be achieved. I don't think it needs to be done at the component level.

PR #128, opened 2022-08-25
Check for nullptr when adding signal

eeberhard:
* Assert that the data pointer given to a signal is not null - if it is null, the MessagePair read / write would fail and the signal would be unusable

* Improve consistency between input and output creating functions

This was a sneaky one. There was a lifecycle component that added an output with uninitialised data pointer, and then initialised the data pointer in the `on_configure_callback`. This meant that the data pointer in the signal message pair was null and not associated with the user-space data pointer. It simply should not be allowed to add a signal with a null data pointer, since it is not associated with any data instance.


domire8 (approved):
Much cleaner now, thanks for the updated checks and messages

PR #129, opened 2022-08-26
Do not publish empty states

eeberhard:
Here's a small feature with potential side-effects. If a state is empty, is there a benefit to publishing? Can we save bandwidth and downstream input logic if we simply skip publishing until the state is non empty? Setting the referenced state variable to empty would then also be a convenient way to temporarily disable automatic publication.

It seems like a nice improvement, but I want to leave this as a draft until we have discussed and thought about the outcomes of this decision.

I posit that an empty publication may still be of interest to a subscriber, and should not be too easily dismissed. A subscription callback for an empty state object could still determine:
1. if anything is connected to the input
2. the rate at which the input is received
3. the name and potentially the reference frame

Let me know your thoughts.


domire8:
Bumping this for finishing the discussion, as requested

eeberhard:
As per our internal discussions, this proposal has been accepted. It is ready for proper review and testing.

domire8 (approved):

domire8 (changes requested):
Clicked too fast haha, the tests need some updates

domire8:
I realized that the python tests are very sparse and don't cover component communication and such. I will add this to the infinity backlog

domire8 (approved):

eeberhard:
Thanks for taking care of those tests!

PR #130, opened 2022-08-27
Rename validate parameter function

domire8:
I currently don't see why this should wait as there are no components under heavy development, and a branch on dynamic-components is ready too.


eeberhard (approved):
Mainly I was waiting because we had some client components in progress, but it's a relatively minor change and is easy enough to patch on the go. Looks good and I'm happy for the improved consistency!

PR #131, opened 2022-08-28
Improve error messages and logs further

domire8:
A previous PR by @eeberhard improved the exception and log messages to be more precise, which I liked a lot, but it wasn't propagated to Python too. This PR makes C++ and Python messages consistent.


eeberhard (approved):
Great stuff 👍 

On the topic of logs for signals; currently, when adding a signal, if the `_topic` parameter already exists, it logs a warning. This is a bit annoying for intentionally dynamic signals (such as the `print_encoded_topic` parameter. Because this is generally a compile-time consequence and not a run-time consequence (e.g. something that the user of the component might cause) I would prefer this to be a debug log.

Properly dealing with dynamic signals (dynamic topic and type) is its own issue that we have already discussed at length, and which I believe should be one of the next main features. In brief, ways to declare signals (and thereby declare their associated parameters) before actually creating the ROS interfaces, and similarly ways to remove ROS interfaces from signals in the case of a dynamic topic / type change (generally during a `cleanup() -> configure()` cycle)

domire8:
Okay do we do this with the next feature, or do you think it should be like that in general, that the add_parameter should only log on debug level if the parameter already exists?

eeberhard:
I think the case of the warn line could be changed here directly, but it's not pressing. It's mainly a question of whether such  warning is useful to the user or to the developer debugging an application. If a `_topic` parameter has already been added for a given signal name is currently entirely baked in to the compilation, so for some components it's a warning that happens every time which the user can do nothing about.

eeberhard (approved):

PR #132, opened 2022-08-31
Implement input callback in Python

domire8:
The next part for this nice feature: The implementation of it in Python with corresponding tests

### DIFF
--- a/source/modulo_components/modulo_components/component_interface.py
+++ b/source/modulo_components/modulo_components/component_interface.py
@@ -390,6 +390,7 @@ class ComponentInterface(Node):
         except (AttributeError, MessageTranslationError) as e:
             self.get_logger().warn(f"Failed to read message for attribute {attribute_name}: {e}",
                                    throttle_duration_sec=1.0)
+            return
         try:
             user_callback()
         except Exception as e:
--- a/source/modulo_components/modulo_components/component_interface.py
+++ b/source/modulo_components/modulo_components/component_interface.py
@@ -503,7 +503,7 @@ class ComponentInterface(Node):
                 raise AddServiceError(f"Service with name '{parsed_service_name}' already exists.")
             signature = inspect.signature(callback)
             if len(signature.parameters) == 0:
-                self.get_logger().error(f"Adding empty service '{parsed_service_name}'.")
+                self.get_logger().debug(f"Adding empty service '{parsed_service_name}'.")
                 service_type = EmptyTrigger
             else:
                 self.get_logger().debug(f"Adding string service '{parsed_service_name}'.")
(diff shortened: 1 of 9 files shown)


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -390,11 +390,14 @@ def __subscription_callback(self, message: MsgT, attribute_name: str, reader: Ca
         except (AttributeError, MessageTranslationError) as e:
             self.get_logger().warn(f"Failed to read message for attribute {attribute_name}: {e}",
                                    throttle_duration_sec=1.0)
+        try:
+            user_callback()
+        except Exception as e:
+            self.get_logger().error(f"Failed to execute user callback in subscription for attribute"
+                                    f" '{attribute_name}': {e}", throttle_duration_sec=1.0)
```
eeberhard, line 397: To be consistent with C++ behaviour, the user callback should not be invoked if the message reading fails. There is a test for this `test_component_interface.cpp`. Could you check that the same case is also tested in Python?

```suggestion
        except (AttributeError, MessageTranslationError) as e:
            self.get_logger().warn(f"Failed to read message for attribute {attribute_name}: {e}",
                                   throttle_duration_sec=1.0)
            return

        try:
            user_callback()
        except Exception as e:
            self.get_logger().error(f"Failed to execute user callback in subscription for attribute"
                                    f" '{attribute_name}': {e}", throttle_duration_sec=1.0)
```

source/modulo_components/test/python/conftest.py
```diff
@@ -1,9 +1,95 @@
...
+    rclpy.init()
+    executor = rclpy.executors.SingleThreadedExecutor()
+    yield executor
+    rclpy.shutdown()
+
+
+@pytest.fixture
+def minimal_cartesian_output(request, random_state):
+    def _make_minimal_cartesian_output(component_type, topic):
+        component = component_type("minimal_cartesian_output", parameter_overrides=[Parameter("period", value=0.1)])
```
eeberhard, line 37: it's unrelated to this PR but I've had this small backlog issue noted down for a while... the default component period in Python is already 0.1 (while in C++ it's 1.0). To be honest I'd prefer to revert the C++ default back to 0.1. I think it's a good balance between being responsive for the user while not being too CPU intensive, and just because 1.0 is a more "pure unit number" I don't think it's a good default period.
```suggestion
        component = component_type("minimal_cartesian_output")
```

eeberhard (commented):
Looks good in general!

source/modulo_components/modulo_components/component_interface.py
domire8: Indeed. While I was adding that test, I realized that the reader for an EncodedState message does not check for the message type, i.e. a component that is expecting a Cartesian state will not give an error in the subscription callback if it receives a joint state..so it will be the same for all other combinations..Do we want this to stay like that?

domire8:
This should now be okay. The user callback will not be executed if the reader raises. The tests have been updated. 

Note that I was using the wrong Future class up until now. Instead of concurrent.future, we should use rclpy.task.Future (I will update dynamic components)

eeberhard (approved):

PR #134, opened 2022-09-12
Use rclpy lifecycle node for lifecycle component

domire8:
I actually don't want to wait too long with showing you this. If you're happy with the changes, I'd merge into humble and then adapt the humble-devel workflow on docker-images to use this branch. This would allow to test these changes at the same time.

The changes itself are quite straight forward, I was able to just write the same functions as in CPP and thanks to the fact that the user only overrides `on_transition_callback`, there are no breaking changes and all tests still pass.


eeberhard (approved):
Nice work! I was trying to think of an elegant way that we could support both galactic and humble; in C++ it could be done with pre-processor macros based on version, but in Python it would be too cumbersome to implement. So, I agree to close it into the humble branch and at some point soon, commit fully to humble

PR #136, opened 2022-09-13
Add option to declare signals

domire8:
As a requested feature, this PR adds the option to declare signals before actually adding them, such that the topic parameter is available on construction and doesn't have to be added manually. I was also able to reduce code duplication by reusing the `declare_signal` function on `add_signal`.

Apart from that there is one small thing that is new: Up until now, it was possible to have an output and input with the same name. I don't think this should be like that and now, a signal with a certain name can either be an input or an output, but not both.


source/modulo_components/include/modulo_components/ComponentInterface.hpp
```diff
@@ -784,17 +808,48 @@ inline void ComponentInterface<NodeT>::set_predicate(
...
   if (this->inputs_.find(parsed_signal_name) != this->inputs_.cend()) {
-    throw exceptions::AddSignalException("Input with name '" + parsed_signal_name + "' already exists.");
+    throw exceptions::AddSignalException("Signal with name '" + parsed_signal_name + "' already exists as input.");
+  }
+  if (this->outputs_.find(parsed_signal_name) != this->outputs_.cend()) {
+    throw exceptions::AddSignalException("Signal with name '" + parsed_signal_name + "' already exists as output.");
+  }
+  std::string topic_name = default_topic.empty() ? "~/" + parsed_signal_name : default_topic;
+  auto parameter_name = parsed_signal_name + "_topic";
+  // TODO is the is_empty condition still valid?
```
domire8, line 828: I'm not sure what to do here, mostly because I can't remember why we put the is_empty condition in the first place. But I think it should be removed, I dont see a reason why we should only set the parameter value if its empty...

eeberhard, line 828: Hmm. `declare_signal` should be done before `add_signal` in the lifecycle case, and manually adding a signal topic parameter should be avoided. Therefore it's unlikely that the parameter would already exist. If it does already exist and has some meaningful value, it definitely shouldn't be set to default (otherwise the sequence of "declare_signal on construction", "user sets topic parameter", "add_signal on configure" would overwrite the user parameter back to default). However, if the parameter exists but is empty, that is a different case.

If someone _really_ wants to manually add the signal topic parameter _and_ leave it empty, then this code would set the signal topic to be the default when the signal is added. I think that's the safest behaviour. The alternative is to not set the parameter value if the parameter already exists, and then in add_input / create_output you'd have to check that the parameter is not empty and the topic name is not empty. It's more work and doesn't bring much apart from the niche feature of "forbidding signals to be created if the parameter hasn't been manually set to be non-empty"

Long story short, let's leave this check in here.

domire8: What about the scenario where someone declares a signal
```
delcare_signal("test"); // parameter test_topic will have value ~/test
```
and then add the signal with a default topic name later on
```
add_signal("test", "this_topic", ...); // parameter test_topic is not empty and will keep value ~/test
```
Do you think that is an unrealistic scenario or just an incorrect usage?

eeberhard: I see that as incorrect usage. There's not a good way of distinguishing if an existing topic parameter value was set by declare_parameter or by a user setting the parameter. Therefore calling add_signal with some different default topic should not apply the new topic. It's a default topic, a.k.a. fallback topic, for the case where the parameter does not exist (or, does not have a value)

domire8:
In the sprint item on Infinity you created @eeberhard, you wrote that upon add signal, any existing signal should be removed and recreated. Here, I do the opposite, i.e. I abort if the signal already exists. Do you have a strong preference there, because I prefer the approach we have now?

eeberhard:
> In the sprint item on Infinity you created @eeberhard, you wrote that upon add signal, any existing signal should be removed and recreated. Here, I do the opposite, i.e. I abort if the signal already exists. Do you have a strong preference there, because I prefer the approach we have now?

Yes, I think I agree with you that `add_signal` should only add a signal, not remove it. Instead we need a dedicated `remove_signal` method that can be called on cleanup or elsewhere. That can be done in another PR

eeberhard (approved):
Looks good to me, I can't spot any problems. It will be nice to see this in action!

PR #137, opened 2022-09-13
Correctly list rclcpp_components as build dependency

domire8:
If we want to avoid having to list `rclcpp_components` as well as `modulo_components` as dependencies of our components, we need to add `build_depend` and `build_export_depend` tags to the package.xml of modulo_components. I could not make it work with just `build_depend`.

Tested with dynamic components on both humble-devel and galactic-devel images. If the build succeeds, it should be fine because the rclcpp_components header is included in the source file and no runtime error should occur (correct me if I'm wrong).


domire8:
It would be great to merge this soon, it allows to test and compare galactic and humble images faster. Moreover we currently have some bugs in dynamic components and backend that I can fix afterwards

buschbapti (approved):

eeberhard (approved):
Yes, this fixes the issue with component registration macro

PR #138, opened 2022-09-15
Correct default value for user callback on add_input

domire8:
Another small issue here. The default value for the user callback was `lambda: None` which doesn't evaluate to False because its a callable, so upon adding an input with a subscription callback, you'd always get the warning that one cannot provide both a subscription callback and a user callback.


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -460,25 +460,31 @@ def add_input(self, signal_name: str, subscription: Union[str, Callable], messag
...
             elif isinstance(subscription, str):
-                if user_callback:
+                if callable(user_callback):
                     signature = inspect.signature(user_callback)
                     if len(signature.parameters) != 0:
                         raise AddSignalError("Provide a user callback that has no input arguments.")
+                else:
+                    if user_callback:
+                        self.get_logger().warn("Provided user callback is not a callable, ignoring it.")
+                    user_callback = lambda: None
```
domire8, line 470: On this line here I get a warning from PEP that you should not assign a lambda expression but use a def. Any preferences? I wouldn't like to define a function somewhere in this file that returns None TBH

eeberhard, line 470: It would be good to avoid PEP warnings. Just define the function object at the lowest scope 
```suggestion
                    def default_callback():
                        return None
                    user_callback = default_callback
```

eeberhard (approved):

buschbapti (approved):

PR #139, opened 2022-09-16
Add option to remove signals

domire8:
As requested, the option to remove a signal (especially for lifecycle components). This required to explicitly reset the ROS publisher/subscription pointers in the Publisher/Subscription handlers as well.

There is still a little question mark if all signals should be removed manually on shutdown


source/modulo_components/modulo_components/component_interface.py
```diff
@@ -330,6 +330,18 @@ def trigger(self, trigger_name: str):
...
         self._publish_predicate(trigger_name)
 
+    def remove_output(self, signal_name):
+        if signal_name not in self._outputs.keys():
+            parsed_signal_name = parse_topic_name(signal_name)
+            if parsed_signal_name not in self._outputs.keys():
+                self.get_logger().debug(f"Unknown output '{signal_name}' (parsed name was '{parsed_signal_name}').")
+                return
+            signal_name = parsed_signal_name
+        if "publisher" in self._outputs[signal_name]:
```
eeberhard, line 340: Is this correct or should it be `self._outputs[signal_name].keys()`?

eeberhard (commented):
Thank you for the continued addition of these features.

For the open question; on further thought I am quite sure we should remove all signals on_shutdown. The component should go into a finalized state with no active interfaces remaining, so it should certainly not respond to input callbacks or any periodic output publications.

source/modulo_components/modulo_components/component_interface.py
domire8: no that is not correct :+1:

domire8:
Following our last discussion, I moved the clear_signals to the shutdown transition. The argumentation was the following:
Cleanup should bring the component back to post-construction state. Given that we don't know if a signal was created on construction or on configure (even though thats something we could do in the future if needed), we cannot just remove all of them, the user will have to remove signals that were added on configure itself. Shutdown however should make sure that any signal is removed.

Even though we don't have the shutdown transition implemented yet in Python on develop branch, I will already ask: Do you know if calling `clear` on a dict in Python will correctly call all destructors of the ROS subscriptions and publishers in our dicts or not? As far as I understand, this is the case in C++.

eeberhard:
>Do you know if calling clear on a dict in Python will correctly call all destructors of the ROS subscriptions and publishers

Python destruction and memory management is a bit different from C++ but it shares the same core concepts.

In C++ when a variable goes out of scope (and is not referenced by anything), it is explicitly destroyed, calling the destructor and de-allocating the memory. In Python, object scope is a little different. When the reference count of a Python object goes to zero, it is not removed immediately but instead is marked as **garbage**; it is the garbage constructor who periodically deletes and unreferenced objects and de-allocates memory. 

Python objects therefore don't have a destructor method, but rather a deleter method. It is possible (and useful) to call the deleter method directly, which handles cleanup logic akin to a C++ destructor, but de-allocation is done by the garbage collector later.

Per your question, calling `dict.clear()` does actually call the deleter on each entry, and therefore clearing the dictionary of signals would also call the deleters on those objects. Unfortunately I see a potential for mismatch or corruption, since any signals we create also get stored internally in the Node `__publishers` and `__subscriptions` lists. So while we would be calling deletion logic on the sub / pub object, we would not be removing all of its references. This is different in C++, since the shared_ptr to the pub / sub is not stored anywhere internally, and it's the caller's responsibility to keep it in scope and prevent destruction. I'm pretty sure this is why they added those `destroy_pub/sub` methods in rclpy, because otherwise the reference management and clean destruction of these interfaces would be very difficult.

It will be the safest to call `remove_signal`, which in turn calls `destroy_pub/sub`, which internally clears it properly in the Node. You could achieve a safe iteration with something like:
```python3
for output in list(self._outputs.keys()):
    self.remove_output(output)
```

Let's leave proper lifecycle state handling in Python for another PR, although it's true we should do it soon (and also to check raise_error and error handling, which is related)

domire8:
Thanks a lot for the clarifications!

I agree to leave lifecycle handling in PR for another PR, but it should be with the humble branch IMO because we can use the lifecycle node there...

eeberhard:
>it should be with the humble branch

Yes absolutely, that is partly what I meant with fully handing the transitions properly with the proper humble framework.

eeberhard (approved):
Good to go!

PR #140, opened 2022-10-03
Fix log level on configure_outputs

domire8:


domire8:
The galactic tests obviously don't pass, the version is unchanged but that's not really required on the humble branch

eeberhard (approved):

PR #145, opened 2022-10-28
2.1.1 release notes

eeberhard:


domire8 (approved):

eeberhard:
In this case, the version hasn't changed, so the checks will not succeed. However I can bypass those requirements for this particular PR

PR #146, opened 2022-11-09
Use rclpy lifecycle node as base class for lifecycle components

domire8:
We should now also merge these changes into develop and get rid of the humble branch. Not sure what the best strategy is though because I had to make an additional commit to move the CHANGELOG entries around. What do you think? Squash and merge or rebase and merge?

@buschbapti


domire8:
It might be easier to squash actually, because I also have to increase the version number and update the CHANGELOG

eeberhard (approved):
I agree you can squash it into one commit #146

PR #147, opened 2022-11-09
Fix test actions and workflows

domire8:
Our test pipeline is still trying to test the code on a `galactic-devel` base image, but this doesn't exist anymore. It's humble now.


eeberhard (approved):

eeberhard:
Although now that I think about it, will the Galactic build and test fail after #146 ? Maybe now is also the time to remove that test entirely

eeberhard:
I was too slow with my realisation :(

domire8:
It's always going to fail yeah but it's not required and if we need to commit on the galactic branch (unlikely) we would still have that test

eeberhard:
Ah that's right, it's not required to pass. I remember the discussion we had a while ago about that.

PR #148, opened 2022-11-10
Remove galactic tests

domire8:
You were right yesterday, we should remove the galactic tests entirely, otherwise it will always seem as if the CI doesn't pass. And if we have changes on the galactic branch, we'll test them locally.


eeberhard (approved):
It's better this way. I initially thought we might support both branches in parallel but I believe we will most likely leave galactic tagged and move on with humble as the next major version

PR #149, opened 2022-11-10
Fix creation of triggers and improve tests

domire8:
This should fix the issue that came up during SRD, that it was not possible to create a trigger on configure. I have honestly no idea why this fix actually works, especially why you could add a trigger on construction but not on configure.


eeberhard (approved):
Thanks for patching and testing that! It makes sense that you don't want to bind a reference to the trigger name, which goes out of scope, but rather just copy the value. 

For why it worked in the constructor, my guess is that the allocation of memory works differently. I'm not going to read too deeply into the specs around this, but binding a reference to a temporary variable during construction may simply be undefined behaviour, while post-construction it's defined (and invalid) behaviour.

PR #150, opened 2022-12-16
Change order of execution in step callback

domire8:
As discussed in [this PR](https://github.com/aica-technology/dynamic-components/pull/201), we should reshuffle the sequence of action in the `step` method of components:

> periodic callbacks, on_step_callback, publish outputs. We can then think of the periodic callbacks as a "pre-step" function


build-server.sh
```diff
@@ -43,6 +43,6 @@ done
 docker pull ghcr.io/aica-technology/ros2-control-libraries:"${BASE_TAG}"
 DOCKER_BUILDKIT=1 docker build -t "${IMAGE_NAME}:${IMAGE_TAG}" "${BUILD_FLAGS[@]}" . || exit 1
 
-if [ "${SERVE_REMOTE}" = true ]; then
+if [ "${SERVE_REMOTE}" == true ]; then
```
buschbapti: This change seems unnecessary https://stackoverflow.com/questions/12948456/is-there-any-difference-between-and-operators-in-bash-or-sh

buschbapti (dismissed):
This seems good

build-server.sh
domire8: What I take from this link is more that I should use double brackets. I still like the double equal sign because its more readable

domire8:
```suggestion
if [[ "${SERVE_REMOTE}" == true ]]; then
```

buschbapti: yeah double brackets in bash is always better

eeberhard (approved):
Thank you, I think this is a sensible change

